PR #3, opened 2024-02-28
fix(UpdateParams/tests) Fix bugs and refine tests to ensure ‘make test’ works

TimmyExogenous:
## Description

When we use the command line to set `ExoCoreLzAppAddress`, the transaction might not be routed correctly because a repeated function is implemented in the withdraw module and the referred module name is misused. This PR deleted the repeat code related to setting `ExoCoreLzAppAddress` in the withdraw module to fix the bug.  
It might be better to merge the deposit and withdrawal modules into one. This also needs to change the code related to precompiles, so we might do it in the future.

The test problem caused by token denominal renaming in the later commit has been resolved.  Additionally, all tests have been refined by using a shared `BaseTestSuite` defined in `testutil/util.go`. The `make test` command can be executed successfully after the second commit. In the subsequent development, we can write clean test code by referring the shared `BaseTestSuite`.

## Todo


adu-web3 (approved):
LGTM

rlan35 (approved):

TimmyExogenous:
The failing "Protobuf / break-check" workflow can be ignored. It reminds us that some protobuf files in the withdraw module have been deleted, but these files have never been used. So, It's fine to delete them. This check won't fail after It's been merged.

MaxMustermann2 (approved):

mikebraver (approved):

PR #4, opened 2024-02-28
feat(dogfood): add module

MaxMustermann2:
[Dogfooding](https://www.notion.so/exocore/Dogfooding-design-96cba225157149d9b5054627bab52275) is the [practice of using one's own products or services](https://en.wikipedia.org/wiki/Eating_your_own_dog_food). The module allows assets deposited on a client chain to be used to secure the Exocore chain. 

- The module uses its parameters to determine the epoch unit (can be hourly, daily or weekly) and the number of such epochs to use for the unbonding period. Besides these two main parameters, the other two parameters are the maximum number of validators and the number of historical entries (required for IBC compatibility).
- Any delegation / undelegation changes and operator opt in and opt outs flow to the consensus engine (through corresponding module hooks) via the ABCI at the end of the epoch in which they are made. This, obviously, requires that the initial set of validators be provided in the `genesis.json` file using a trust-minimized (verifiable) approach which is done via the bootstrapping contract.
- The unbonding time is calculated as the number of unbonding epochs * duration of an epoch after this time.
- The module is designed to act as a drop-in replacement for the Cosmos SDK's staking module, and implements most of the interfaces required by other modules in the file `impl_sdk.go`.

Note that this module is not responsible for distributing rewards to the operators and delegators; it will have to be created separately.

Supersedes ExocoreNetwork/exocore-private#114 from the older repo.


TimmyExogenous (approved):
LGTM. It looks like the PR has already been reviewed in the old repo.

mikebraver (approved):

rlan35 (approved):

PR #5, opened 2024-02-28
chore(ci): fix labeler and consensuswarn

MaxMustermann2:
This PR fixes the failing `consensuswarn` and `labeler` workflows.


MaxMustermann2:
The labeler workflow runs off of the base branch and will not be triggered in this PR. We will have to commit it blindly.

MaxMustermann2:
> The labeler workflow runs off of the base branch and will not be triggered in this PR. We will have to commit it blindly.

The same is true for the consensus warn workflow, which is running on the old root containing the '.BaseApp' string.

PR #8, opened 2024-03-01
chore(ci): re-attempt fix

MaxMustermann2:
This is another attempt to fix all of the Github workflows for the `develop` branch. Since all of our PRs will merge to develop, the base branch must have its workflows passing successfully before our changes can have their code quality assessed.

Noteworthy changes:
- I have disabled some of the parts or even full workflows which aren't applicable for our repository, since they were failing otherwise.
- The dependency check (by Github) only fails if the vulnerability is rated at least `high`.
- The patch version of 2 dependencies was upgraded.
- All of the `proto` files have been linted and the `go` files re-generated.


MaxMustermann2:
The failing "Protobuf / break-check" workflow, in my opinion, should be ignored for this pull request because the changes are only cosmetic in nature (the Go code remains unchanged). However, do note that the JSON serialization of the objects will change because their JSON name has actually changed. Per my understanding, we do not use JSON serialization outside of the genesis file or the API or the CLI; so we should be good from a chain running perspective. The staking portal could need some changes.

"Consensus Warn / main (pull_request_target)" runs on the base (default?) branch and thus will not succeed until the `DeliverTx` code reaches that branch.

TimmyExogenous (approved):
LGTM

MaxMustermann2:
If it is necessary to do so, the old JSON serialization can be retained with the `json_name` field. However, I would prefer to follow the recommended convention and make the changes within the consumers of the JSON API like the staking portal instead.

rlan35 (approved):

mikebraver (approved):
LGTM

MaxMustermann2:
Since the JSON API change will not be immediately deployed to our nodes, I will send a draft PR to the staking portal to account for it. The deployment will have to be coordinated amongst the two.

PR #10, opened 2024-03-04
chore(ci): fix super linter config

MaxMustermann2:
The super-linter.yml workflow seems to be running on pull requests and branches. However, for pull requests, it fails silently with the logs reporting an error. On branches, it fails actually with an ❌ emoji.


MaxMustermann2:
The failing lint check here is, again, the protobuf's `break-check`. There is a singular non-comment change within the proto files, which is fixing a typo of a field in an unused module. Hence, it can be ignored.

TimmyExogenous (approved):
LGTM

PR #11, opened 2024-03-05
feat(dogfood): asset ids, oracle usage

MaxMustermann2:
As #4 stands today, the dogfood module relies on operator and delegation hooks to decide the validator updates which need to be passed to Tendermint. However, in a multi-asset staking situation, this is not feasible because the price of the asset (which is being delegated / undelegated) may change during the epoch. Hence, a function called `GetAvgDelegatedValue` is envisaged in the `restaking_assets_manage` keeper. This function should return the sum of the currently delegated values (not pending undelegation) multiplied by the average price of the assets within the epoch (derived from the oracle module). With this new design, a lot of the complexities of the previous design have been eliminated or offloaded to other modules; however, it is a heavier operation because of the iteration of validators that is required at each epoch.


x/dogfood/keeper/abci.go
```diff
@@ -1,12 +1,21 @@
...
 
 func (k Keeper) EndBlock(ctx sdk.Context) []abci.ValidatorUpdate {
+	id, _ := k.getValidatorSetID(ctx, ctx.BlockHeight())
+	if !k.IsEpochEnd(ctx) {
+		// save the same id for the next block height.
+		k.setValidatorSetID(ctx, ctx.BlockHeight()+1, id)
+		return []abci.ValidatorUpdate{}
+	}
+	defer k.ClearEpochEnd(ctx)
 	// start with clearing the hold on the undelegations.
```
TimmyExogenous, line 19: I haven't seen the code to increase the undelegationHoldCount. Why does it need to be decreased here? Additionally, I have looked at the related code about OnHold in Cosmos-SDK and Evmos, It's a great feature. But I didn't find this feature being used in Cosmos-SDK and Evmos.

TimmyExogenous (approved):
LGTM

x/dogfood/keeper/abci.go
MaxMustermann2: > I haven't seen the code to increase the undelegationHoldCount. Why does it need to be decreased here? 

You are right; I made a mistake. The line that increases it was accidentally removed. I will add it back. Great catch!

> Additionally, I have looked at the related code about OnHold in Cosmos-SDK and Evmos, It's a great feature. But I didn't find this feature being used in Cosmos-SDK and Evmos.

Evmos certainly does not use it, as far as I can tell. Cosmos-SDK introduced the feature to support interchain-security, which does use it. I do not think any standalone chains use it.

rlan35 (approved):

mikebraver (approved):

PR #15, opened 2024-03-12
fix(ci): semgrep, cov-comment, buf-break-no-fail

MaxMustermann2:
In the last push to the develop branch, semgrep [complained](https://github.com/ExocoreNetwork/exocore/actions/runs/8243299667/job/22543646196) about minor things: docker compose security attributes (fixed), a mutex not being unlocked (which has now been added to the list ignored files) and the conversion of uint64 to uint32 (fix taken from 26d7b72efd7b7d5c028bc015079d7d016590bddd).

An edit has been made to the `test.yml` workflow by splitting the coverage and comment workflows into two parts, thereby fixing a [potential security hole](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/).

Lastly, this PR silences the failure of buf's breaking check instead posting it as a comment to the PR. Since we are in the phase of rapid prototyping and testing, these failures add unneeded noise to all of the pull requests. However, the information provided by the workflow is still relevant, and hence, its status is posted as a comment to the pull request instead.


testutil/network/network.go
```diff
@@ -218,6 +218,7 @@ func New(l Logger, baseDir string, cfg Config) (*Network, error) {
 	// only one caller/test can create and use a network at a time
 	l.Log("acquiring test network lock")
 	lock.Lock()
+	defer lock.Unlock()
```
TimmyExogenous: This might deviate from the original purpose of this function. What it aims to do is to unlock through `Cleanup` by the caller. It's a global lock. 
I also encountered this error when I ran `semgrep` locally. I am considering whether we should delete the  `testutil/network` directory since we don't use it now. or add the `lock.Unlock` in this function to eliminate the error, just like you did.

MaxMustermann2: I have added the `testutil` folder to `.semgrepignore`.

MaxMustermann2:
Currently, there is a change in `proto/exocore/dogfood/v1/dogfood.proto` reported by the break-check action as a breaking change. However, the run has passed since we set `continue-on-error` to true. The result of the workflow has been uploaded as an artifact. The artifact contains the PR number as well.

There is a downside to this approach that the result is not immediately visible on the PR's front page (although it is visible within the files changed tab). To work around this problem, I set up another workflow to comment its result. When  `proto-comment.yml` is merged to `develop` it will ideally start running in response to (and after) `proto.yml` automatically.

image

TimmyExogenous (approved):
Great work.

adu-web3 (approved):
LGTM 👍

PR #21, opened 2024-03-25
[feat] implement the operator interfaces expected by dogfood

TimmyExogenous:
# Description

This PR implements the operator interfaces expected by the dogfood module. Additionally, it solves a problem encountered when running `make proto-download-deps`.

The makefile about `proto-download-deps` downloads the code from branch `cosmos-sdk\origin\master`. However, we currently rely on the historical version, Therefore we need to update the makefile to download the corresponding version of the dependency proto code.

The implemented operator interfaces are as follows:
 * jail/unjail
 * IsOperatorJailedForChainID
 * GetActiveOperatorsForChainID
 * GetAvgDelegatedValue
 * GetOperatorAssetValue
 * SlashWithInfractionReason

# Todo
 * The slash and power interfaces expected by the dogfood module
 * The missed calling in OptInToCosmosChain
 * Refactor the design related to the power store
 * The implementation related to self-delegation and `min_self_delegation`


proto/exocore/operator/v1/tx.proto
```diff
@@ -55,6 +64,8 @@ message OptedInfo {
   uint64 opted_in_height = 2;
   // opted_out_height is the exocore block height at which the operator opted out
   uint64 opted_out_height = 3;
+  // jailed defined whether the operator has been jailed from bonded status or not.
+  bool jailed = 4;
```
MaxMustermann2, line 62: Design question: Should jailing be done on a per-AVS basis or globally? We consider two potential reasons for slashing: first being the downtime and second being malicious activity like double-signing.

If an operator is slashed for downtime, we should only jail them for that specific AVS.
If an operator is slashed for malicious activity, we should ban them on all AVS. 

What do you think?

proto/exocore/operator/v1/tx.proto
```diff
@@ -45,6 +46,14 @@ message OperatorInfo {
   string operator_meta_info = 3;
   // client_chain_earning_addr_list is the client chain earning address list.
   ClientChainEarningAddrList client_chain_earnings_addr = 4;
+  // commission defines the commission parameters.
+  cosmos.staking.v1beta1.Commission commission = 5 [(gogoproto.nullable) = false, (amino.dont_omitempty) = true];
+  // min_self_delegation is the validator's self declared minimum self delegation.
+  string min_self_delegation = 6 [
```
MaxMustermann2: Shouldn't MinSelfDelegation be a property of the AVS and not of the operator?

proto/exocore/operator/v1/tx.proto
TimmyExogenous: Both the malicious activities and slash proportions are configured by the specific Avs. If we ban them on all AVS when an operator acts maliciously on one Avs,  Is it a problem that low-level misconduct will cause a maximum punishment?
I noticed a tombstone mechanism to mitigate the impact of initially likely categories of non-malicious protocol faults in cosmos-sdk. So, I'm also unsure about the appropriate level for the slash mechanism in Exocore.

proto/exocore/operator/v1/tx.proto
TimmyExogenous: In the cosmos-SDK, this is configured by the validator, the delegator can choose the validator according to this. So I add this field in the operator. It's property to configure a MinSelfDelegation in the AVS parameter to set a threshold for opt-in.
Do you think operators still need to have a MinSelfDelegation that must be greater than or equal to the MinSelfDelegation specified in the AVS parameters in order to attract delegators?

proto/exocore/operator/v1/tx.proto
MaxMustermann2: > If we ban them on all AVS when an operator acts maliciously on one Avs, Is it a problem that low-level misconduct will cause a maximum punishment?

Yes, as discussed, we should only jail temporarily for low level misconduct.

>  I noticed a tombstone mechanism to mitigate the impact of initially likely categories of non-malicious protocol faults in cosmos-sdk. So, I'm also unsure about the appropriate level for the slash mechanism in Exocore.

I might be misinterpreting this but I think they have two things:
- Jailing, which is temporary or permanent. An operator is jailed for downtime and slashed a small amount. Unjailing is first on the protocol level (upon expiry of the jailing time), and then the operator must send a transaction indicating that they have fixed the misconfiguration that led to the slashing time.
- Tombstoning, which is permanent (equivalent to freezing). An operator is tombstoned for double signing, slashed a larger amount and permanently jailed. The only "mitigation" offered by this mechanism is that a validator cannot be slashed twice if they double sign across multiple blocks, even if evidence is received for each of those blocks.

I observed that permanent jailing is equivalent to tombstoning (and the SDK even performs them together). So we can remove that feature. Either an operator is jailed temporarily on a specific AVS, or they are permanently blocked from them all. For ease of use, we can replace the word "tombstone" with "freeze" like in our proposed design.

proto/exocore/operator/v1/tx.proto
MaxMustermann2: There are two things here:
- Operator-level minimum self delegation, which is used by the operator to attract delegations. If the operator's self-delegation falls below this level, they are kicked out of the validator set. If we are to implement this, we should implement this in USD value as well. If, at the end of an epoch, the self-delegated value falls below the USD amount, the operator is not considered for inclusion in the validator set for any AVS.
- AVS-level minimum self-delegation, which is used by the AVS to ensure its operators have some skin in the game. This, too, must be denominated in USD (unfortunately). This is because we intend to support restaking on multiple chains, and validators from chain A (say Ethereum) will not necessarily be validators on chain B (say Avalanche). If we denominate the minimum self-stake in native restaked ETH and separately in native restaked AVAX, a validator active only on Ethereum but not on Avalanche will not be able to come up with any native restaked AVAX to meet the minimum. We can apply this mechanism similar to the operator-level implementation. If, at the end of an epoch, the self-delegated value falls below the USD amount required by that AVS, the operator is not considered for inclusion in the validator set for that AVS.

x/operator/keeper/operator.go
```diff
@@ -54,7 +54,30 @@ func (k *Keeper) IsOperator(ctx sdk.Context, addr sdk.AccAddress) bool {
...
+	k.cdc.MustUnmarshal(value, info)
+	// call the handleFunc
+	handleFunc(info)
+	// restore the info after handling
+	bz := k.cdc.MustMarshal(info)
+	store.Set(infoKey, bz)
+	return nil
+}
+
+func (k *Keeper) SetOptedInfo(ctx sdk.Context, operatorAddr, avsAddr string, info *operatortypes.OptedInfo) error {
```
bwhour, line 93: In the SetOptedInfo function, in addition to verifying the address format,  is it necessary to verify that the **info** parameter contains all the necessary information required to join the AVS and that this information conforms to the predefined rules and constraints ？

bwhour: and also  SetOperatorInfo function.

x/dogfood/keeper/impl_sdk.go
```diff
@@ -103,7 +96,7 @@ func (k Keeper) SlashWithInfractionReason(
 	}
 	// TODO(mm): add list of assets to be slashed (and not just all of them).
 	// based on yet to be finalized slashing design.
-	return k.slashingKeeper.SlashWithInfractionReason(
+	return k.operatorKeeper.SlashWithInfractionReason(
 		ctx, accAddress, infractionHeight,
 		power, slashFactor, infraction,
```
bwhour, line 102: lack of validation of legitimacy of some parameters (slashFactor --> slashProportion), need we  make sure it is in a reasonable  range ？

x/operator/keeper/avs_operator_shares.go
```diff
@@ -335,3 +335,35 @@ func (k *Keeper) GetStakerShare(ctx sdk.Context, avsAddr, stakerID, operatorAddr
...
+			return nil, err
+		}
+		// truncate the USD value to int64
+		ret = append(ret, share.TruncateInt64())
+	}
+	return ret, nil
+}
+
+func (k *Keeper) GetOperatorAssetValue(ctx sdk.Context, operator sdk.AccAddress, chainID string) (int64, error) {
+	avsAddr, err := k.avsKeeper.GetAvsAddrByChainID(ctx, chainID)
```
bwhour, line 359: Is it necessary to validate the validity of the formatting rules for avsAddr and operatorAddr except for null?

- UpdateOperatorShare DeleteOperatorShare GetOperatorShare UpdateAVSShare
- UpdateStateForAsset DeleteAssetState (assetID, avsAddr)
- BatchUpdateShareForAVSAndOperator Does this function require validation of key-value pairs in the mapping avsOperatorChange?

bwhour (changes requested):
pls check the comment

proto/exocore/operator/v1/tx.proto
TimmyExogenous: Yeah, this needs to be considered in the design and implementation of the slash module.

proto/exocore/operator/v1/tx.proto
TimmyExogenous: As we discussed, we can only retain the `MinSelfDelegation` of AVS.

x/operator/keeper/operator.go
TimmyExogenous: It seems we don't need to consider this for now. Both of these interfaces are modifications to the KV store at the low level. If we need to verify the information, it might be best to determine which parameters need checking and perform the parameter validation at the entry point of the call.

x/dogfood/keeper/impl_sdk.go
TimmyExogenous: The `slashFactor` is a parameter of the slashing module of cosmos-sdk. It should have been verified when configuring it.

x/operator/keeper/avs_operator_shares.go
TimmyExogenous: I will add code in the `optIn` function to check if the AVS address is an EVM contract address. Other checks are unnecessary because all entry points of calls will check the operator address type. If the type is incorrect, the relevant operator and AVS cannot be queried in the KV store. As for `avsOperatorChange`, value checks will be handled uniformly in `UpdateAssetDecValue`.

bwhour (approved):

MaxMustermann2 (approved):

PR #22, opened 2024-03-26
fix(app): remove unused modules

MaxMustermann2:
The following modules, which were inherited from the prior code base, have been removed from the app's entry point and associated tests.

- incentives: This module is used to provide incentives to certain projects, via governance, on a chain. Since our chain is not designed to be a home to many projects, this module is not needed.
- inflation: This is the replacement of the Cosmos SDK's `x/mint` module in the inherited code. It also relies on the ratio of currently bonded tokens, which is meaningless for our chain. Hence, it has been removed. It will be replaced by the `x/exomint` module in a separate pull request.
- vesting: As the name suggests, this is a token vesting module. It allows for the addition of new vesting accounts and setting their parameters, and their conversion to a real account when vesting is complete. This module is not critical to the functioning of the chain and has been removed. It may be added back later, if deemed necessary.
- revenue: This module is used for sharing the gas fee with the smart contract developer. It has been removed for the same reason as that of the incentives module.
- claims: This is the airdrop module and is unused at this point in time. The recovery and the ERC20 modules used to depend on the claims module before ExocoreNetwork/evmos#3 (which needs to be merged before this PR). That dependency has been removed, and therefore the recovery module and the ERC20 modules can survive without the claims module.
- distribution: This module handles the distribution of native token rewards amongst the delegators. Since our delegators do not have a real native address, this module cannot function. The tasks of this module should be performed by the reward module.

Note that the recovery module and the ERC20 modules are still required, with the latter depending on the parameters of the former. The recovery module is used to recover any assets that are mistakenly sent to the Cosmos derived address instead of the Ethereum derived address over IBC. The ERC20 module helps in the conversion of IBC assets to ERC20 assets and vice-versa.


MaxMustermann2:
The " Consensus Warn / main (pull_request_target) Failing after 50s " is expected and acceptable. It is happening because we are removing the unused modules.

MaxMustermann2:
~The distribution module is still pending removal. I am converting to draft until I figure out how to do it.~

The distribution module has been removed by removing the claims module (which needs the distribution module) first. The claims module has been removed by removing it as a dependency from the recovery and the ERC20 modules.

app/export.go
```diff
@@ -116,141 +107,52 @@ func (app *ExocoreApp) ExportAppStateAndValidators(
...
-	allowedAddrsMap := make(map[string]bool)
-
-	for _, addr := range jailAllowedAddrs {
-		_, err := sdk.ValAddressFromBech32(addr)
-		if err != nil {
-			return err
-		}
-		allowedAddrsMap[addr] = true
-	}
+func (app *ExocoreApp) prepForZeroHeightGenesis(
```
TimmyExogenous, line 103: Do you know if the function of exporting genesis at zero height is still needed? Will any situation use this function?

x/evm/types/params.go
```diff
@@ -11,9 +11,7 @@ var (
 	DefaultEVMDenom               = utils.BaseDenom
 	ExocoreAvailableEVMExtensions = []string{
 		"0x0000000000000000000000000000000000000800", // Staking precompile
-		"0x0000000000000000000000000000000000000801", // Distribution precompile
```
TimmyExogenous, line 14: Should we modify the addresses of other precompile contracts to prevent any addresses from being left vacant after removing the distribution and vesting modules?

app/app.go
```diff
@@ -936,35 +1023,44 @@
...
 	app.sm = module.NewSimulationManagerFromAppModules(app.mm.Modules, overrideModules)
 
-	autocliv1.RegisterQueryServer(app.GRPCQueryRouter(), runtimeservices.NewAutoCLIQueryService(app.mm.Modules))
+	autocliv1.RegisterQueryServer(
+		app.GRPCQueryRouter(),
+		runtimeservices.NewAutoCLIQueryService(app.mm.Modules),
+	)
 
 	reflectionSvc, err := runtimeservices.NewReflectionService()
 	if err != nil {
```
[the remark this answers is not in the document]
MaxMustermann2: The " Consensus Warn / main (pull_request_target) Failing after 50s " is expected and acceptable. It is happening because we are removing the unused modules.

TimmyExogenous (approved):

app/export.go
MaxMustermann2: This function is used in `ExportAppStateAndValidators`, which is a function in the `AppI` interface required by Cosmos SDK. It is used for simulation purposes.

Within the evmos fork of the Cosmos SDK, the `forZeroHeight` variable is always false. So it does not look like this function is used at the moment.

However, we should retain it anway - for when we do implement and use the fuzz testing simulation framework.

x/evm/types/params.go
MaxMustermann2: There is no downside in leaving vacant addresses. Although I do plan to add an operator precompile at `801`.

app/export.go
```diff
@@ -57,19 +54,12 @@ func NewDefaultGenesisState(cdc codec.Codec) simapp.GenesisState {
...
-	cdc.MustUnmarshalJSON(rawGenesis, &inflationGenesis)
-	inflationGenesis.Params.MintDenom = utils.BaseDenom
-	defaultGenesis[inflationtypes.ModuleName] = cdc.MustMarshalJSON(&inflationGenesis)
-
 	// claims module
-	claimsGenesis := claimstypes.GenesisState{}
-	rawGenesis = defaultGenesis[claimstypes.ModuleName]
-	cdc.MustUnmarshalJSON(rawGenesis, &claimsGenesis)
-	claimsGenesis.Params.ClaimsDenom = utils.BaseDenom
-	defaultGenesis[claimstypes.ModuleName] = cdc.MustMarshalJSON(&claimsGenesis)
```
bwhour, line 72: pls remove all these comment lines once it's unused.

go.mod
```diff
@@ -215,14 +215,12 @@ replace (
...
-	//github.com/cosmos/cosmos-sdk => ../cosmos-sdk
 	//fix cosmos-sdk error
 	github.com/cosmos/gogoproto => github.com/cosmos/gogoproto v1.4.10
 	// use Evmos geth fork
 	github.com/ethereum/go-ethereum => github.com/evmos/go-ethereum v1.10.26-evmos-rc2
-	// use exocore fork of evmos
-	github.com/evmos/evmos/v14 => github.com/ExocoreNetwork/evmos/v14 v14.1.1-0.20240205024453-5e8090e42ef4
-	//github.com/evmos/evmos/v14 => ../ExocoreNetwork/evmos
+	// use exocore fork of evmos TODO
+	github.com/evmos/evmos/v14 => github.com/MaxMustermann2/evmos/v14 v14.0.0-20240326155756-4f464426e972
```
bwhour: once evmos merged, this part and related need to be changed.

precompiles/deposit/deposit_test.go
```diff
@@ -60,11 +58,11 @@ func (s *DepositPrecompileSuite) TestRunDepositTo() {
 	opAmount := big.NewInt(100)
 	assetAddr := usdtAddress
 	commonMalleate := func() (common.Address, []byte) {
-		valAddr, err := sdk.ValAddressFromBech32(s.Validators[0].OperatorAddress)
-		s.Require().NoError(err)
-		val, _ := s.App.StakingKeeper.GetValidator(s.Ctx, valAddr)
-		coins := sdk.NewCoins(sdk.NewCoin(utils.BaseDenom, sdk.NewInt(1e18)))
-		s.App.DistrKeeper.AllocateTokensToValidator(s.Ctx, val, sdk.NewDecCoinsFromCoins(coins...))
+		// valAddr, err := sdk.ValAddressFromBech32(s.Validators[0].OperatorAddress)
```
bwhour: pls remove the comment lines instead of comment

precompiles/delegation/delegation_test.go
```diff
@@ -340,11 +339,11 @@ func (s *DelegationPrecompileSuite) TestRunUnDelegateFromThroughClientChain() {
 	}
 	commonMalleate := func() (common.Address, []byte) {
 		// prepare the call input for delegation test
-		valAddr, err := sdk.ValAddressFromBech32(s.Validators[0].OperatorAddress)
-		s.Require().NoError(err)
-		val, _ := s.App.StakingKeeper.GetValidator(s.Ctx, valAddr)
-		coins := sdk.NewCoins(sdk.NewCoin(utils.BaseDenom, sdk.NewInt(1e18)))
-		s.App.DistrKeeper.AllocateTokensToValidator(s.Ctx, val, sdk.NewDecCoinsFromCoins(coins...))
+		// valAddr, err := sdk.ValAddressFromBech32(s.Validators[0].OperatorAddress)
+		// s.Require().NoError(err)
```
bwhour: pls remove the comment lines instead of comment

precompiles/delegation/delegation_test.go
```diff
@@ -98,11 +97,11 @@ func (s *DelegationPrecompileSuite) TestRunDelegateToThroughClientChain() {
 	}
 	commonMalleate := func() (common.Address, []byte) {
 		// prepare the call input for delegation test
-		valAddr, err := sdk.ValAddressFromBech32(s.Validators[0].OperatorAddress)
-		s.Require().NoError(err)
-		val, _ := s.App.StakingKeeper.GetValidator(s.Ctx, valAddr)
-		coins := sdk.NewCoins(sdk.NewCoin(utils.BaseDenom, sdk.NewInt(1e18)))
-		s.App.DistrKeeper.AllocateTokensToValidator(s.Ctx, val, sdk.NewDecCoinsFromCoins(coins...))
+		// valAddr, err := sdk.ValAddressFromBech32(s.Validators[0].OperatorAddress)
```
bwhour: pls remove the comment lines instead of comment

PR #23, opened 2024-04-01
feat(multiple): read from genesis file

MaxMustermann2:
This pull request sets up genesis data structures for the modules `assets`, `delegation`, `operator` and `dogfood`. The data is read, marshaled and then validated before the node starts. Once the node starts, the data is sent to the module keepers at genesis where stateful validations can take place.

Secondly, it makes some minor changes (comments and clean-up) in the `dogfood` module.

Lastly, it also sets up a CLI way for validators to export their consensus public key as a bytes32 object to use within Solidity.

```bash
exocored keys consensus-pubkey-to-bytes --output json | jq -r .bytes
F0F6919E522C5B97DB2C8255BFF743F9DFDDD7AD9FC37CB0C1670B480D0F9914
```

For reference, the (JSON) string object can be exported as follows:

```bash
exocored tendermint show-validator
{"@type":"/cosmos.crypto.ed25519.PubKey","key":"8PaRnlIsW5fbLIJVv/dD+d/d162fw3ywwWcLSA0PmRQ="}
```

The same key can be recovered from the mnemonic (which can also be read from stdin manually)
```bash
echo "${MNEMONIC}" | exocored init localnet --chain-id exocorelocal_233-1 --recover
```


MaxMustermann2:
Closing this PR to break it down into multiple PRs.

PR #24, opened 2024-04-02
[R4R] feat(oracle)

leonz789:
## Description

Oracle module is used to collect external prices from sources(offchain/onchain) outside of exocorechain. And the prices access served by oracle module will be used by other modules like operator for voting power calculation.
Supersedes #6 , reconstruct the commits for easy to review.

commits:
1. *proto file definition and some ignite scalffold dependencies*
2. *mainly generated files from proto definition*
3. **implement keeper for state**
4. *add tests for keeper*
5. *add cache serves as memory storage for price collection across blocks*
6. **add aggregator, this implements the main logic for price aggregation**
Screenshot 2024-04-03 at 03 48 43


MaxMustermann2:
Can you please fix the failing workflows?

bwhour (approved):
reviewed

x/oracle/keeper/aggregator/aggregator.go
bwhour, line 1: The status field uses an integer type to indicate status (1 for open, 2 for closed). Using enumerated types or constant definitions instead of magic numbers makes the code easier to understand and maintain.

x/oracle/keeper/aggregator/aggregator_aggregator.go
bwhour, line 1: The **AggregatorContext** contains multiple mappings (e.g. **validatorsPower, rounds, aggregators**) that may be accessed and modified concurrently in a multi-threaded environment. If this is the case, you need to ensure that access to these mappings is thread-safe and may need to use mutex locking or other concurrency control mechanisms

x/oracle/keeper/aggregator/aggregator_calculator.go
bwhour, line 1: len(r.prices) < cap(r.prices)
Its a good practice taking  takes into account not exceeding the preset capacity.
Is it necessary to take into account other boundary conditions, such as validation of the input parameters ?

bwhour, line 1: Concurrency control needs to be considered when multiple goroutines may access or modify the **deterministicSource** mapping in the calculator.  it is a common requirement to handle concurrency in applications like blockchain and price aggregation.

x/oracle/keeper/aggregator/aggregator_filter.go
bwhour, line 1: If there exists a pSource with no prices, accessing pSource.Prices[0] will trigger panic

bwhour, line 1: Use string concatenation (validator + strconv. Itoa(int(pSource.SourceId))) to generate the mapped key, whether this key generation rule may cause key conflicts

bwhour, line 1: For collection operations (Add operations), assuming that they always succeed without checking the return value can hide problems.

x/oracle/keeper/aggregator/worker.go
bwhour, line 1: Setting f, c, and a to nil in the **seal method** may help with garbage collection, but if there are operations in these objects that require explicit shutdown or cleanup of resources, extra care is needed to ensure that the resources are freed properly

x/oracle/module.go
```diff
@@ -0,0 +1,201 @@
...
+// AppModule
+// ----------------------------------------------------------------------------
+
+// AppModule implements the AppModule interface that defines the inter-dependent methods that modules need to implement
+type AppModule struct {
+	AppModuleBasic
+
+	keeper        keeper.Keeper
+	accountKeeper types.AccountKeeper
+	bankKeeper    types.BankKeeper
```
MaxMustermann2, line 99: Can you add a comment here to say that they are only used in the simulation?

x/oracle/keeper/keeper.go
```diff
@@ -0,0 +1,73 @@
...
+	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
+)
+
+type (
+	Keeper struct {
+		cdc           codec.BinaryCodec
+		storeKey      storetypes.StoreKey
+		memKey        storetypes.StoreKey
+		paramstore    paramtypes.Subspace
+		stakingKeeper stakingkeeper.Keeper
```
MaxMustermann2, line 28: Please convert this to the minimum required interface in `expected_keepers.go`.

x/oracle/keeper/aggregator/worker.go
bwhour, line 1: _list4Calculator, list4Aggregator := w.f.filtrate(msg)_: if the **filtrate method** encounters any exceptions while processing a message (e.g., the message is formatted incorrectly), the current implementation will not catch or handle those exceptions

x/oracle/keeper/keeper.go
MaxMustermann2, line 28: Please convert this to the minimum required interface in `expected_keepers.go`.

x/oracle/module.go
MaxMustermann2, line 100: Can you add a comment here to say that they are only used in the simulation?

MaxMustermann2 (changes requested):
Not fully reviewed.

x/oracle/keeper/cache/caches.go
bwhour, line 1: im.PSources = append(im.PSources, item.PSources...)
This case is If there are **new PSources** to be added, it will add the entire PSources array to the existing entry, not just the new SourceId. If  if the new PSources array contains some new and some old (already existing in the current entry) SourceIds, then the price sources associated with the old SourceId will be added repeatedly; over time, if new PSources are added frequently, the PSources array in each entry may become very large, affecting data processing and query efficiency。

bwhour, line 1: functions in this cache file seems no error handling and logging logic.

x/oracle/keeper/prices.go
```diff
@@ -0,0 +1,189 @@
...
+		} else if prevTokenId != tokenId && price.TokenId > 0 {
+			list = append(list, price)
+			prevTokenId = tokenId
+			price = types.Prices{TokenId: tokenId}
+		}
+		if nextRoundId {
+			price.NextRountId = binary.BigEndian.Uint64(iterator.Value())
+		} else {
+			var val types.PriceWithTimeAndRound
+			k.cdc.Unmarshal(iterator.Value(), &val)
```
bwhour, line 94: k.cdc.Unmarshal has error return which need to be handle. can k.cdc.MustMarshal ?
image

x/oracle/keeper/prices.go
```diff
@@ -0,0 +1,189 @@
...
+
+func (k Keeper) GetPriceTRRoundId(ctx sdk.Context, tokenId int32, roundId uint64) (price types.PriceWithTimeAndRound, found bool) {
+	store := k.getPriceTRStore(ctx, tokenId)
+
+	b := store.Get(types.PricesRoundKey(roundId))
+	if b == nil {
+		return
+	}
+
+	k.cdc.Unmarshal(b, &price)
```
bwhour, line 125: k.cdc.Unmarshal has error return which need to be handle. can k.cdc.MustMarshal ?

x/oracle/keeper/prices.go
```diff
@@ -0,0 +1,189 @@
...
+	store := k.getPriceTRStore(ctx, tokenId)
+	nextRoundIdB := store.Get(types.PricesNextRountIdKey)
+	if nextRoundIdB == nil {
+		return
+	}
+	nextRoundId := binary.BigEndian.Uint64(nextRoundIdB)
+	b := store.Get(types.PricesRoundKey(nextRoundId - 1))
+	if b != nil {
+		//should always be true
+		k.cdc.Unmarshal(b, &price)
```
bwhour, line 142: k.cdc.Unmarshal has error return which need to be handle. can k.cdc.MustMarshal ?

x/oracle/keeper/aggregator/aggregator_calculator.go
```diff
@@ -0,0 +1,168 @@
...
+			item.power = new(big.Int).Add(item.power, pw.power)
+			updated = true
+			if common.ExceedsThreshold(item.power, totalPower) {
+				r.price = item.price
+				confirmed = true
+			}
+			return
+		}
+	}
+	if len(r.prices) < cap(r.prices) {
```
bwhour, line 50: len(r.prices) < cap(r.prices)
Its a good practice taking  takes into account not exceeding the preset capacity.
Is it necessary to take into account other boundary conditions, such as validation of the input parameters ?

x/oracle/keeper/aggregator/aggregator_calculator.go
```diff
@@ -0,0 +1,168 @@
...
+	detId     string
+	prices    []*priceAndPower
+	price     *big.Int
+	timestamp string
+	// confirmed bool
+}
+
+// udpate priceAndPower for a specific DSRoundID, if the price exists, increase its power with provided data
+// return confirmed=true, when detect power exceeds the threshold
+func (r *roundPrices) updatePriceAndPower(pw *priceAndPower, totalPower *big.Int) (updated bool, confirmed bool) {
```
bwhour, line 34: In the updatePriceAndPower method, when new prices and weights are added to roundPrices, if any exceptions are encountered (e.g., if pw.price or pw.power is nil), the code doesn't perform error handling

leonz789:
Merging with the original branch introduced confusion due to too many commits. I will reconstruct it to create a cleaner version. Apologies for the mess. (will follow this 6-commit structure for easy review)

leonz789:
I will close this pr, and #30 will take place for the oracle module feature pr.

PR #25, opened 2024-04-03
feat(assets): load genesis state

MaxMustermann2:
The `x/assets` module should load the genesis state from disk for the network to successfully bootstrap with the rPOS process. This pull request makes that happen.

The genesis state is structured as follows:

```json
{
"params": {...}
"client_chains": [{...}, {...}, {...}],
"tokens": [{...},{...}],
"deposits": [
    {
        "staker": "stakerID_lzIDhex",
        "deposits": [
            {
                 "asset_id": "assetAddress_lzIDhex",
                 "info": {....}
            }, {...}
        ]
    }, {...}
]}
```

At genesis, the module should load:

- The client chains
- The client chain tokens (with 0 deposit value, since deposits are handled in the next bullet item)
- The deposits (delegated and available-to-delegate both) by each staker for each registered token (or assumed 0 otherwise).

The stateless validation follows this algorithm:

- There are no duplicate client chains.
- There are no tokens for which we don't have a client chain.
- The basic info for a token (everything but the amount deposited) is not nil.
- The tokens have an Ethereum-style case-insensitive hex address (this check could be removed in the future). From this address, we derive the `assetID` which is always in lower case.
- There are no duplicate entries for the tokens.
- The token entries have no deposit value (but the staker entries may, which are added to the 0 value for that asset).
- The token supply is not nil and is positive.

For each staker deposit entry, we check:

- The `stakerID` is valid: lowercase, exactly one underscore, with the left side of the underscore non-empty and a hex address, and the right side a hex uint64.
- There are no duplicate entries for a `stakerID`

Within each list of deposits indexed by `assetID` per `stakerID`, we check for each `assetID`:

- There are no duplicate entries for an `assetID` within the list of deposits for said `stakerID`.
- The `assetID` is registered previously in the tokens list.
- All the values (deposited, withdrawable and pending unbonding) are supplied
- The amount pending unbonding is zero (which is true for a genesis chain).
- The supplied amounts of deposit and withdrawable values are not negative.
- The withdrawable amount is less than or equal to the deposit amount.
- The deposit amount does not exceed the max supply of the token previously provided.


x/assets/keeper/app_chain.go
TimmyExogenous, line 1: Should we move this file to the AVS module? Seems like it provides functions for the app-chain AVS.

TimmyExogenous (approved):
Great work. LGTM

x/assets/keeper/genesis.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/ExocoreNetwork/exocore/x/assets/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// InitGenesis initializes the module's state from a provided genesis state.
+func (k Keeper) InitGenesis(ctx sdk.Context, data *types.GenesisState) {
+	if err := k.SetParams(ctx, &data.Params); err != nil {
+		panic(err)
+	}
+	// client_chain.go
```
bwhour, line 13: Each time the loop iterates, the infoCopy variable is assigned to the info variable, and this assignment operation actually creates a copy of each element in the data.ClientChains or data.Tokens arrays. If these arrays contain a large number of elements, or if the elements themselves take up a large amount of memory, then this copy operation can result in significant memory overhead

x/assets/keeper/genesis.go
```diff
@@ -0,0 +1,59 @@
+package keeper
+
+import (
+	"github.com/ExocoreNetwork/exocore/x/assets/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// InitGenesis initializes the module's state from a provided genesis state.
+func (k Keeper) InitGenesis(ctx sdk.Context, data *types.GenesisState) {
```
bwhour, line 9: Traversing and working with large amounts of data in InitGenesis (e.g. customer chain information, token information, and deposit information) may lead to performance issues, especially during initial block loading

x/assets/keeper/app_chain.go
MaxMustermann2: Agreed, we should do that within the AVS PR or afterwards.

x/assets/keeper/genesis.go
MaxMustermann2: Added a `TODO` to this effect for later.

bwhour (approved):

PR #26, opened 2024-04-03
 feat(delegation): load genesis state

MaxMustermann2:
This PR subsumes #27 within itself, and therefore should be merged after that one. For ease of review, this PR may be reviewed after the merge of #27 happens and this PR is rebased.

The `x/delegation` module needs to load the genesis state of delegations (for each possible combination of staker, asset and operator) for the network to successfully bootstrap with the rPOS process.

The genesis state is structured as follows.

```json
{
    "delegations": [
        {
            "staker_id": "0xaaaaaa_0x65",
            "delegations": [
                {
                    "asset_id": "0xbbbb_0x65",
                    "per_operator_amounts": [
                        {"key": "exo1.....", "value": {"amount": "700"}},
                        {"key": "exo1.....", "value": {"amount": "500"}}, {...}
                    ]
                }, {...}
            ]
        }, {...}
    ]
}
```

At genesis, the module should perform these delegations as it would if the chain were live. To do this, we call the exact same `DelegateTo` function that is called by the precompile and use a flag to skip (1) the operator frozen check, (2) the operator module update and (3) the delegation hooks.

The stateless validation uses this algorithm:
-  There are no repeat stakers.
- The `stakerID` is valid: lowercase, exactly one underscore, with the left side of the underscore non-empty and a hex address, and the right side a hex `uint64`.
- For each `stakerID`, there is no repeating `assetID`.
- Each `assetID` is valid (defined above).
- The `LayerZeroID` derived from the `stakerID` equals that derived from the `assetID` for each provided `stakerID` + `assetID` combination.
- The per-operator amount, for each operator, is non-nil and non-negative.
- The provided operator address is bech32 encoded appropriately.


TimmyExogenous (approved):
LGTM

x/delegation/keeper/genesis.go
```diff
@@ -0,0 +1,59 @@
...
+func (k Keeper) InitGenesis(
+	ctx sdk.Context,
+	gs delegationtype.GenesisState,
+) []abci.ValidatorUpdate {
+	for _, level1 := range gs.Delegations {
+		stakerID := level1.StakerID
+		// #nosec G703 // already validated
+		stakerAddress, lzID, _ := assetstype.ParseID(stakerID)
+		// we have checked IsHexAddress already
+		stakerAddressBytes := common.HexToAddress(stakerAddress)
```
bwhour, line 23: If the genesis state contains a lot of delegate information, the current nested loop may affect the performance of the initialization. Consider whether it is possible to improve efficiency by optimizing data structures or parallel processing.

1. Consider distributing processing operations on each delegate to different goroutines to execute in parallel. Ensure data consistency and thread safety when processing in parallel.Pay attention to the number of goroutines to avoid resource contention caused by creating too many goroutines
2. If the delegateTo function supports batch processing, that is, processing multiple delegates at a time, we can collect a batch of delegates in each goroutine and call delegateTo to process them. This reduces the number of function calls and increases efficiency.

x/delegation/types/genesis.go
```diff
@@ -0,0 +1,102 @@
...
+
+// DefaultGenesis returns the default genesis state
+func DefaultGenesis() *GenesisState {
+	return NewGenesis([]DelegationsByStaker{})
+}
+
+// Validate performs basic genesis state validation returning an error upon any
+// failure.
+func (gs GenesisState) Validate() error {
+	stakers := make(map[string]struct{}, len(gs.Delegations))
```
bwhour, line 30: During the verification process, for each level of data (stakeholder, asset, operator), a separate hash table is used to check for uniqueness. Be aware that memory usage may increase when processing large amounts of data.

bwhour (approved):

x/delegation/keeper/genesis.go
MaxMustermann2: ~1. It is not trivial to build something like this because the order of `goroutine` execution is not guaranteed. This can result in inconsistency across nodes in calculating the state root, since it depends on the order of insertion.~
2. Within `delegateTo`, there are 6 different function calls, only 4 of which can be batched either on the staker + asset level or on the operator + asset level. Others cannot be batched, hence, batching is not necessarily possible.

I have added a TODO for this.

x/delegation/types/genesis.go
MaxMustermann2: As discussed, added a TODO for this.

PR #27, opened 2024-04-03
feat(operator): load genesis state

MaxMustermann2:
This PR subsumes #25 within itself, and therefore should be merged after that one. For ease of review, this PR may be reviewed after the merge of #25 happens and this PR is rebased.

The `x/operator` module should load the genesis state from disk for the network to successfully bootstrap with the rPOS process. This pull request makes that happen.

The genesis state is structured as follows:

```json
{
    "operators": [
          {...}, {...}
    ],
    "operator_records": [
        {
            "operator_address": "exo1...",
            "chains": [
                {
                    "chain_id": "exocoretestnet_233-1",
                    "consensus_key": "0xFFFFAAA...."
                }, {...}
             ]
        }, {...}
    ]
}
```

At genesis, the module should:
- Register all of the operators
- Opt them into the chains (for which the consensus key is provided)
- Store the provided consensus keys against the operator address + `chainID` combination

The stateful validations performed are the same as the validations on a live chain, since all the calls are routed to those very functions.

The stateless validations are listed below.
- The operator addresses in `operators` list are of the correct bech32 format.
- The client chain earnings addresses, if provided, have unique LayerZero chain ids (for that operator), and the value of those addresses is an Ethereum hex-address (TODO: make this format dependent on the `lzID`).
- The operator addresses in `operator_records` list are correctly bech32 encoded.
- The operator in `operator_records` exists in `operators`.
- The provided consensus keys are unique for the `chainID`.
- The consensus keys can be parsed from hex (Solidity's bytes32).


x/operator/keeper/genesis.go
```diff
@@ -0,0 +1,47 @@
...
+			// then set pub key
+			if err := k.setOperatorConsKeyForChainID(
+				ctx, operatorAddr, detail.ChainID, key, true,
+			); err != nil {
+				panic(err)
+			}
+			bootstrapping = bootstrapping || ctx.ChainID() == detail.ChainID
+		}
+		if !bootstrapping {
+			// TODO: consider removing this check
```
TimmyExogenous: Should we consider multiple app chains here? It appears that this could lead to panics, especially if the inputs involve multiple app chains. Would it be simpler and more effective to only support the Exocore chain here? Then the other app chain AVSs could be launched after Exocore has been launched

TimmyExogenous (approved):
LGTM

x/operator/keeper/genesis.go
MaxMustermann2: I see your point. I don't feel too strongly to keep this `panic` at this point, however, I added it to keep it future proof.

The `InitGenesis` function is designed (by the Cosmos-SDK) to support actual genesis as well as chain restarts and upgrades (that are not happening via governance). In the last two cases, it may be possible that app chains are imported to the state from the JSON file.

What do you think?

MaxMustermann2: Actually, this panic should not be here at all. It is not the responsibility of the operator module to validate this. It is something done by the dogfood module. I will move that check there.

x/operator/types/genesis.go
```diff
@@ -0,0 +1,126 @@
...
+			for _, info := range op.ClientChainEarningsAddr.EarningInfoList {
+				lzID := info.LzClientChainID
+				if _, found := lzIDs[lzID]; found {
+					return errorsmod.Wrapf(
+						ErrInvalidGenesisData,
+						"duplicate lz client chain id %d", lzID,
+					)
+				}
+				lzIDs[lzID] = struct{}{}
+				// TODO: consider removing this check for non-EVM chains
```
bwhour: To support non-EVM chains, we can introduce a new field or parameter that identifies the type of client chain and select different validation logic depending on the type of chain

x/operator/types/errors.go
```diff
@@ -3,33 +3,72 @@ package types
 import errorsmod "cosmossdk.io/errors"
 
 var (
-	ErrNoKeyInTheStore = errorsmod.Register(ModuleName, 0, "there is not the key for in the store")
+	ErrNoKeyInTheStore = errorsmod.Register(
+		ModuleName, 0,
+		"there is not the key for in the store",
```
bwhour: the "for" seems redundant

bwhour (approved):

PR #30, opened 2024-04-09
[R4R] feat(oracle)

leonz789:
## Description

Oracle module is used to collect external prices from sources(offchain/onchain) outside of exocorechain. And the prices access served by oracle module will be used by other modules like operator for voting power calculation.
Supersedes #6 , reconstruct the commits for easy to review.

commits:
1. *proto file definition and some ignite scalffold dependencies*
2. *mainly generated files from proto definition*
3. **implement keeper for state**
4. *add tests for keeper*
5. *add cache serves as memory storage for price collection across blocks*
6. **add aggregator, this implements the main logic for price aggregation**
Screenshot 2024-04-03 at 03 48 43


leonz789:
ref: #24

x/oracle/keeper/prices.go
```diff
@@ -0,0 +1,189 @@
...
+		found = true
+	}
+	return
+}
+
+func (k Keeper) GetNextRoundId(ctx sdk.Context, tokenId int32) (nextRoundId uint64) {
+	nextRoundId = 1
+	//store := getPriceTRStore(ctx, k.storeKey, tokenId)
+	store := k.getPriceTRStore(ctx, tokenId)
+	nextRoundIdB := store.Get(types.PricesNextRountIdKey)
```
bwhour: NextRountIdKey ---> **NextRoundIdKey**  , a lot of alike spelling error

x/oracle/keeper/prices.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+//func(k Keeper) SetPriceTR(ctx sdk.Context, tokenId int32, priceTR){}
+
+func (k Keeper) GetPriceTRRoundId(ctx sdk.Context, tokenId int32, roundId uint64) (price types.PriceWithTimeAndRound, found bool) {
+	store := k.getPriceTRStore(ctx, tokenId)
+
+	b := store.Get(types.PricesRoundKey(roundId))
+	if b == nil {
+		return
```
bwhour, line 111: return uninitialized variables

x/oracle/keeper/prices.go
```diff
@@ -0,0 +1,189 @@
...
+	return
+}
+
+func (k Keeper) GetPriceTRLatest(ctx sdk.Context, tokenId int32) (price types.PriceWithTimeAndRound, found bool) {
+	//	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefix(types.PricesKeyPrefix))
+	//	store = prefix.NewStore(store, types.PricesKey(tokenId))
+	store := k.getPriceTRStore(ctx, tokenId)
+	nextRoundIdB := store.Get(types.PricesNextRountIdKey)
+	if nextRoundIdB == nil {
+		return
```
bwhour, line 125: return uninitialized variables

x/oracle/keeper/prices.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	if price.TokenId > 0 {
+		list = append(list, price)
+	}
+	return
+}
+
+func (k Keeper) AppendPriceTR(ctx sdk.Context, tokenId int32, priceTR types.PriceWithTimeAndRound) {
+	nextRoundId := k.GetNextRoundId(ctx, tokenId)
+	if nextRoundId != priceTR.RoundId {
```
bwhour: Return an error instead of failing silently        
If the round ID of the price information you are trying to add is not the expected next round ID, the caller will receive a clear error message indicating the problem

leonz789:
Thanks @bwhour and @MaxMustermann2  for the pr review(#24 ), here's former comments i just copied here and corresponding replies:

1. workflows fail: 
A: working

2. currency concern for maps:
A: currently there's no concurrent access, will update when we do the concurrency related performance update.

3. 'If there exists a pSource with no prices, accessing pSource.Prices[0] will trigger panic'
   A: before the message be processed, it has been checked to make sure the pSource at least has one price

4. 'Use string concatenation (validator + strconv. Itoa(int(pSource.SourceId))) to generate the mapped key, whether this key generation rule may cause key conflicts'
   A: won't be, it's like a uri each level will not be conflict under parent scope

5. '*list4Calculator, list4Aggregator := w.f.filtrate(msg)*: if the **filtrate method** encounters any exceptions while processing a message (e.g., the message is formatted incorrectly), the current implementation will not catch or handle those exceptions'
   A: The format check is and should be done in the first place when the message start processing(have been done in the `checkmsg`)

6. 'In the updatePriceAndPower method, when new prices and weights are added to roundPrices, if any exceptions are encountered (e.g., if pw.price or pw.power is nil), the code doesn't perform error handling'
   A: same as above
7. `len(r.prices) < cap(r.prices)
   Its a good practice taking takes into account not exceeding the preset capacity.
   Is it necessary to take into account other boundary conditions, such as validation of the input parameters ?`
   A: this length check is meant to make sure we don't accept values more than the limitation we set, just use the capicity as a boundary limitation
  
8. '*list4Calculator, list4Aggregator := w.f.filtrate(msg)*: if the **filtrate method** encounters any exceptions while processing a message (e.g., the message is formatted incorrectly), the current implementation will not catch or handle those exceptions'
   A: Format check is/should be done in the checkmsg and antehandler

9. 'im.PSources = append(im.PSources, item.PSources...)
   This case is If there are **new PSources** to be added, it will add the entire PSources array to the existing entry, not just the new SourceId. If if the new PSources array contains some new and some old (already existing in the current entry) SourceIds, then the price sources associated with the old SourceId will be added repeatedly; over time, if new PSources are added frequently, the PSources array in each entry may become very large, affecting data processing and query efficiency'
   A: the cache is just used as temporary storage, we wanna it be fast, and for the case you mentioned here, the duplicated srouceIds has been checked and ignored before put into cache.

10. 'For collection operations (Add operations), assuming that they always succeed without checking the return value can hide problems.'
 A: not sure what this meant for.

Others:
1. error handling and logs
 A: has added some and updated with new commits, will check further

2. Update status to enum instead of integer
 A: will update

x/oracle/keeper/prices.go
qinglin89: expected, found will be initialized to false by default, just return the `false` is expected.

x/oracle/keeper/prices.go
qinglin89: This method will be invoked in endblock. It won't respond to the caller directly, but we could certainly add some return info for the log.

x/oracle/keeper/prices.go
qinglin89: expected

x/oracle/keeper/prices.go
qinglin89: udpated

proto/exocore/oracle/info.proto
```diff
@@ -0,0 +1,40 @@
+syntax = "proto3";
+package exocore.oracle;
+
+option go_package = "github.com/ExocoreNetwork/exocore/x/oracle/types";
+
+message Chain{
```
TimmyExogenous, line 9: Should we use the chains and assets defined and registered in the assets module instead of defining them anew here?

proto/exocore/oracle/info.proto
```diff
@@ -0,0 +1,40 @@
...
+  //eg."bitcoin"
+  string name = 1;
+  //TODO: metadata
+  string desc = 2;
+}
+
+message Token{
```
TimmyExogenous, line 17: Same comment as above

proto/exocore/oracle/query.proto
```diff
@@ -0,0 +1,157 @@
...
+import "exocore/oracle/validator_update_block.proto";
+import "exocore/oracle/index_recent_params.proto";
+import "exocore/oracle/index_recent_msg.proto";
+import "exocore/oracle/recent_msg.proto";
+import "exocore/oracle/recent_params.proto";
+
+option go_package = "github.com/ExocoreNetwork/exocore/x/oracle/types";
+
+// Query defines the gRPC querier service.
+service Query {
```
TimmyExogenous, line 19: Should it provide an RPC that only returns the recent price of a specified asset?   Seems like the `Prices`  includes a list of `PriceWithTimeAndRound`.  perhaps the other modules only need a return type as below:
```go
type Price struct {
	Value   sdkmath.Int
	Decimal uint8
}
```

leonz789:
For the workflow: Tests / test-unit-cover (pull_request) 
it seems that the workflow can't handle the 'monkey patch' well, the test itself is actually passed.

x/oracle/keeper/prices.go
leonz789: expected, found will be initialized to false by default, just return the false is expected.

x/oracle/keeper/prices.go
leonz789: expected

proto/exocore/oracle/recent_params.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
+package exocore.oracle;
+
+import "exocore/oracle/params.proto";
+
+option go_package = "github.com/ExocoreNetwork/exocore/x/oracle/types";
+
+// RecentParams represents the params cached for recent blocks
+message RecentParams {
```
TimmyExogenous, line 9: Seems like the process of updating params is the same as updating prices, right?

proto/exocore/oracle/token_feeder.proto
```diff
@@ -0,0 +1,45 @@
...
+  uint64 minimum = 2;
+}
+
+//specify data from which source is needed
+//rule_1: specified sources
+//rule_2: n out of total sources are required
+message RuleSource{
+  //refer to params.sourceList.ID, when length>0, ignore the other field, when 1st set to 0, means all valid sources, 
+  // length==0->check next field:minimum
+  repeated uint64 source_ids = 1 [(gogoproto.customname) = "SourceIDs"];
```
TimmyExogenous, line 22: why the `source_ids` list is also defined here, there is already a list in `NOMSource`, and it the second field of `RuleSource`.

proto/exocore/oracle/token_feeder.proto
```diff
@@ -0,0 +1,45 @@
...
+message TokenFeeder{
+  //refer to params.tokenList, from 1
+  uint64 token_id = 1 [(gogoproto.customname) = "TokenID"];
+  //refer to params.ruleList, 0 means no restriction, accept any source including customer defined
+  uint64 rule_id = 2 [(gogoproto.customname) = "RuleID"];
+  //include, from 1, when some token's feeder had been stop and then restart, 
+  // the token_id will be continuous from previous one
+  uint64 start_round_id = 3 [(gogoproto.customname) = "StartRoundID"];
+  //include, first block which start_round_id can be settled is at least start_base_block+1
+  uint64 start_base_block = 4;
```
TimmyExogenous, line 37: Is it a block height of the Exocore chain?

proto/exocore/oracle/tx.proto
```diff
@@ -0,0 +1,53 @@
...
+  string creator = 1 [(cosmos_proto.scalar) = "cosmos.ValidatorAddressString"];
+  //refer to id from Params.TokenFeeders, 0 is reserved, invalid to use
+  uint64 feeder_id = 2 [(gogoproto.customname) = "FeederID"];
+
+  // prices price with its corresponding source
+  repeated PriceSource prices = 3;
+  //on which block commit does this message be built on
+  uint64 based_block = 4;
+  // nonce represents the unique number to disginguish duplicated messages
+  int32 nonce = 5;
```
TimmyExogenous, line 33: Why does the nonce need to be defined in the msg, isn't there a default nonce mechanism in the system?

bwhour (approved):

tools/tools.go
TimmyExogenous, line 1: What's the usage of this file?

x/oracle/keeper/aggregator/aggregator.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+type roundInfo struct {
+	// this round of price will start from block basedBlock+1, the basedBlock served as a trigger to notify validators to submit prices
+	basedBlock uint64
+	// next round id of the price oracle service, price with thie id will be record on block basedBlock+1 if all prices submitted by validators(for v1, validators serve as oracle nodes) get to consensus immedately
+	nextRoundID uint64
+	// indicate if this round is open for collecting prices or closed in either condition that success with a consensused price or not
+	// 1: open, 2: closed
+	status int32
```
TimmyExogenous: It's better to use an enum type.

x/oracle/keeper/common/types.go
```diff
@@ -0,0 +1,174 @@
...
+import (
+	"errors"
+	"math/big"
+	"sort"
+
+	"github.com/ExocoreNetwork/exocore/x/oracle/types"
+)
+
+const (
+	// maxNonce indicates how many messages a validator can submit in a single roudn to offer price
```
TimmyExogenous: correct `roudn` to round.

x/oracle/keeper/single.go
```diff
@@ -0,0 +1,146 @@
...
+	agc = aggregator.NewAggregatorContext()
+	if ok := recacheAggregatorContext(ctx, agc, k, c); !ok {
+		// this is the very first time oracle has been started, fill relalted info as initialization
+		initAggregatorContext(ctx, agc, k, c)
+	}
+	return agc
+}
+
+func recacheAggregatorContext(ctx sdk.Context, agc *aggregator.AggregatorContext, k Keeper, c *cache.Cache) bool {
+	from := ctx.BlockHeight() - common.MaxNonce
```
TimmyExogenous, line 43: The nonce represents a block duration, right? Why use nonce naming it? Or is it that for transactions involving price updates, only one transaction is allowed per validator within a block so that the nonce can represent one block interval exactly?

x/oracle/keeper/aggregator/aggregator_aggregator.go
TimmyExogenous, line 1: What's the meaning of the file name `aggregator_aggregator`? It's a little bit confused with the `aggregator.go`, could we use a better name?

x/oracle/keeper/aggregator/aggregator.go
```diff
@@ -0,0 +1,267 @@
...
+
+type PriceItemKV struct {
+	TokenID uint64
+	PriceTR types.PriceTimeRound
+}
+
+type roundInfo struct {
+	// this round of price will start from block basedBlock+1, the basedBlock served as a trigger to notify validators to submit prices
+	basedBlock uint64
+	// next round id of the price oracle service, price with thie id will be record on block basedBlock+1 if all prices submitted by validators(for v1, validators serve as oracle nodes) get to consensus immedately
```
TimmyExogenous: wrong spelling, `thie` and `immedately`

TimmyExogenous (commented):
There is a lot of code in this PR, so didn't review it very deeply, just left some questions.

proto/exocore/oracle/genesis.proto
```diff
@@ -0,0 +1,35 @@
...
+// GenesisState defines the oracle module's genesis state.
+message GenesisState {
+  // module params
+  Params params     = 1 [(gogoproto.nullable) = false];
+  // prices of all tokens
+  repeated Prices prices_list = 2 [(gogoproto.nullable) = false];
+  
+  //TODO: userDefinedTokenFeeder
+  // latest block on which the validator set be updated
+  ValidatorUpdateBlock validator_update_block = 4;
```
mikebraver: Should this start from 3?

proto/exocore/oracle/recent_params.proto
leonz789: not much, the recentParams is mainly used to track the modification of tokenFeeder

proto/exocore/oracle/info.proto
leonz789: Yes, the integration related with other modules will be raise as new PR.

proto/exocore/oracle/info.proto
leonz789: Will update in new PR

proto/exocore/oracle/query.proto
leonz789: Will raise new PR for the integration with other modules

proto/exocore/oracle/token_feeder.proto
leonz789: yes

proto/exocore/oracle/tx.proto
leonz789: this is used to limit the number of transactions/prices a validator can submit for one round

tools/tools.go
leonz789: this is used by ignite

x/oracle/keeper/single.go
leonz789: `nonce` is used to limit the number of transactions/prices one validator could submit in one round for a specific tokenFeeder

x/oracle/keeper/aggregator/aggregator.go
leonz789: corrected

x/oracle/keeper/aggregator/aggregator_aggregator.go
leonz789: file names updated

x/oracle/keeper/common/types.go
leonz789: corrected

proto/exocore/oracle/token_feeder.proto
leonz789: Yes, i think these two fields can be merged into one single filed, will update in a new PR

proto/exocore/oracle/info.proto
TimmyExogenous: I see.

TimmyExogenous (approved):

proto/exocore/oracle/genesis.proto
leonz789: corrected

app/app.go
```diff
@@ -448,15 +455,12 @@ func NewExocoreApp(
 		operatorTypes.StoreKey,
 		avsManagerTypes.StoreKey,
 		avsTaskTypes.StoreKey,
+		oracleTypes.StoreKey,
 	)
 
 	// Add the EVM transient store key
```
MaxMustermann2: Update or remove this comment.

app/app.go
```diff
@@ -977,6 +985,7 @@ func NewExocoreApp(
 		withdrawTypes.ModuleName,
 		rewardTypes.ModuleName,
 		exoslashTypes.ModuleName,
+		oracleTypes.ModuleName,
```
MaxMustermann2, line 988: Please explain the rationale for these orders, if there is any. Or add a comment anyway if there is no particular rationale.

app/app.go
```diff
@@ -1338,9 +1347,8 @@ func initParamsKeeper(
 	paramsKeeper.Subspace(ibcexported.ModuleName)
 	paramsKeeper.Subspace(icahosttypes.SubModuleName)
 	// ethermint subspaces
-	paramsKeeper.Subspace(evmtypes.ModuleName).
-		// nolint: staticcheck
-		WithKeyTable(evmtypes.ParamKeyTable())
+	paramsKeeper.Subspace(evmtypes.ModuleName).WithKeyTable(evmtypes.ParamKeyTable()) //nolint:staticcheck
+	paramsKeeper.Subspace(oracleTypes.ModuleName).WithKeyTable(oracleTypes.ParamKeyTable())
```
MaxMustermann2, line 1351: Shouldn't this also have the same `nolint` line ?

proto/exocore/oracle/price.proto
```diff
@@ -0,0 +1,45 @@
...
+// token price with timestamp fetched from source
+// {price:"12345",decimal:"2"}->price: 123.45 usdt
+message PriceTimeDetID {
+  // price at a specific point(timestamp of non-deterministic source, roundId of deteministic source)
+  string price = 1;
+  // decimal of the corresponding price 
+  int32 decimal = 2;
+  // timestamp when the price corresponding to
+  string timestamp = 3;
+  // det_id is used for deterministic source to tell of which round from this soure the price is corresponded
```
MaxMustermann2: typo: `soure`

tools/tools.go
```diff
@@ -0,0 +1,11 @@
...
+
+package tools
+
+import (
+	_ "github.com/cosmos/gogoproto/protoc-gen-gocosmos"
+	_ "github.com/golang/protobuf/protoc-gen-go"
+	_ "github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway"
+	_ "github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger"
+	_ "github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2"
+)
```
MaxMustermann2, line 11: Nice!

x/oracle/client/cli/tx.go
```diff
@@ -0,0 +1,35 @@
...
+	"time"
+
+	"github.com/spf13/cobra"
+
+	"github.com/cosmos/cosmos-sdk/client"
+	// "github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/ExocoreNetwork/exocore/x/oracle/types"
+)
+
+var DefaultRelativePacketTimeoutTimestamp = uint64((time.Duration(10) * time.Minute).Nanoseconds())
```
MaxMustermann2, line 14: nit: please remove

x/oracle/client/cli/tx.go
```diff
@@ -0,0 +1,35 @@
...
+
+// const (
+//	flagPacketTimeoutTimestamp = "packet-timeout-timestamp"
+//	listSeparator              = ","
+//)
```
MaxMustermann2, line 19: nit: please remove

x/oracle/client/cli/tx_create_price.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/cosmos/cosmos-sdk/client/tx"
+	"github.com/spf13/cobra"
+)
+
+var _ = strconv.Itoa(0)
+
+func CmdCreatePrice() *cobra.Command {
+	cmd := &cobra.Command{
+		// TODO: support v1 single sourceID for temporary
+		Use:   "create-price feederid basedblock nonce sourceid decimal price timestamp detid optinoal(price timestamp detid) optional(desc)",
```
MaxMustermann2, line 19: typo: `optinoal`

x/oracle/genesis.go
```diff
@@ -0,0 +1,65 @@
...
+	// Get all indexRecentMsg
+	indexRecentMsg, found := k.GetIndexRecentMsg(ctx)
+	if found {
+		genesis.IndexRecentMsg = &indexRecentMsg
+	}
+	genesis.RecentMsgList = k.GetAllRecentMsg(ctx)
+	genesis.RecentParamsList = k.GetAllRecentParams(ctx)
+	// this line is used by starport scaffolding # genesis/module/export
+
+	return genesis
```
MaxMustermann2, line 64: Good call making this function initially

proto/exocore/oracle/tx.proto
MaxMustermann2, line 33: Is there another way besides introducing another `nonce` parameter ?

x/oracle/keeper/common/types.go
```diff
@@ -0,0 +1,174 @@
...
+
+	// these two threshold value used to set the threshold to tell when the price had come to consensus and was able to get a final price of that round
+	ThresholdA = 2
+	ThresholdB = 3
+
+	// maxDetId each validator can submit, so the calculator can cache maximum of maxDetId*count(validators) values, this is for resistance of malicious validator submmiting invalid detId
+	MaxDetID = 5
+
+	// consensus mode: v1: as soon as possbile
+	Mode = 1
```
MaxMustermann2, line 24: Enum please

x/oracle/keeper/aggregator/context.go
```diff
@@ -0,0 +1,267 @@
...
+	// 1. for sealed worker, the KVStore has been updated
+	for feederID, round := range agc.rounds {
+		if round.status == 1 {
+			feeder := agc.params.GetTokenFeeder(feederID)
+			// TODO: for mode=1, we don't do aggregate() here, since if it donesn't success in the transaction execution stage, it won't success here
+			// but it's not always the same for other modes, switch modes
+			switch common.Mode {
+			case 1:
+				expired := feeder.EndBlock > 0 && uint64(ctx.BlockHeight()) >= feeder.EndBlock
+				outOfWindow := uint64(ctx.BlockHeight())-round.basedBlock >= uint64(common.MaxNonce)
```
MaxMustermann2, line 168: Can you explain the correlation between height change and maximum nonce?

x/oracle/module.go
```diff
@@ -0,0 +1,219 @@
...
+// ConsensusVersion is a sequence number for state-breaking change of the module. It should be incremented on each consensus-breaking change introduced by the module. To avoid wrong/empty versions, the initial version should be set to 1
+func (AppModule) ConsensusVersion() uint64 { return 1 }
+
+// BeginBlock contains the logic that is automatically triggered at the beginning of each block
+func (am AppModule) BeginBlock(_ sdk.Context, _ abci.RequestBeginBlock) {}
+
+// EndBlock contains the logic that is automatically triggered at the end of each block
+func (am AppModule) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) []abci.ValidatorUpdate {
+	cs := keeper.GetCaches()
+	validatorUpdates := am.keeper.GetValidatorUpdates(ctx)
```
MaxMustermann2, line 158: Instead of relying on this, a `hook` approach could potentially work out better. We can add it to the `dogfood` module to inform any interested modules when the validator set changes (which happens during `EndBlock` under certain conditions).

x/oracle/keeper/aggregator/context.go
```diff
@@ -0,0 +1,267 @@
...
+
+		feederIDUint64 := uint64(feederID)
+		round := agc.rounds[feederIDUint64]
+		if round == nil {
+			round = &roundInfo{
+				basedBlock:  latestBasedblock,
+				nextRoundID: latestNextRoundID,
+			}
+			if left >= common.MaxNonce {
+				round.status = 2
```
MaxMustermann2, line 220: Why is this done?

x/oracle/module.go
```diff
@@ -0,0 +1,219 @@
...
+		}
+		logger.Info(logInfo)
+		ctx.EventManager().EmitEvent(event)
+	}
+	// TODO: emit events for success sealed rounds(could ignore for v1)
+
+	logger.Info("prepare for next oracle round of each tokenFeeder")
+	agc.PrepareRound(ctx, 0)
+
+	cs.CommitCache(ctx, true, am.keeper)
```
MaxMustermann2, line 217: Maybe I missed this but I don't see `c.validators.commit(ctx, k)` being called as a result of this function even though `GetValidatorUpdates` might have a non-0 length.

x/oracle/keeper/keeper.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+type (
+	Keeper struct {
+		cdc        codec.BinaryCodec
+		storeKey   storetypes.StoreKey
+		memKey     storetypes.StoreKey
+		paramstore paramtypes.Subspace
+		authority  string
+		common.KeeperStaking
```
MaxMustermann2, line 23: why embed it?

proto/exocore/oracle/tx.proto
MaxMustermann2: Also why use `int32` and not unsigned ?

app/app.go
```diff
@@ -666,6 +670,7 @@ func NewExocoreApp(
 	app.ExoSlashKeeper = slashKeeper.NewKeeper(appCodec, keys[exoslashTypes.StoreKey], app.AssetsKeeper)
 	app.AVSManagerKeeper = *avsManagerKeeper.NewKeeper(appCodec, keys[avsManagerTypes.StoreKey], &app.OperatorKeeper, app.AssetsKeeper)
 	app.TaskKeeper = avsTaskKeeper.NewKeeper(appCodec, keys[avsTaskTypes.StoreKey], app.AVSManagerKeeper)
+	app.OracleKeeper = oracleKeeper.NewKeeper(appCodec, keys[oracleTypes.StoreKey], memKeys[oracleTypes.MemStoreKey], app.GetSubspace(oracleTypes.ModuleName), app.StakingKeeper)
```
MaxMustermann2, line 672: Should this be added to `app.go` without integration with other modules?

proto/exocore/oracle/recent_params.proto
```diff
@@ -0,0 +1,15 @@
...
+  // block height of which the params from
+  uint64 block = 1; 
+  // params the module params
+  Params params = 2;
```
MaxMustermann2, line 13: Where is this used?

proto/exocore/oracle/token_feeder.proto
```diff
@@ -0,0 +1,45 @@
...
+  repeated uint64 source_ids = 1 [(gogoproto.customname) = "SourceIDs"];  
+  //n out of total sources are required
+  NOMSource nom = 2;
+}
+
+//Tokenfeeder represents a price oracle for one token
+message TokenFeeder{
+  //refer to params.tokenList, from 1
+  uint64 token_id = 1 [(gogoproto.customname) = "TokenID"];
+  //refer to params.ruleList, 0 means no restriction, accept any source including customer defined
```
MaxMustermann2, line 31: customer ?

proto/exocore/oracle/token_feeder.proto
```diff
@@ -0,0 +1,45 @@
...
+  uint64 start_round_id = 3 [(gogoproto.customname) = "StartRoundID"];
+  //include, first block which start_round_id can be settled is at least start_base_block+1
+  uint64 start_base_block = 4;
+  //set as count of blocks, for how many blocks interval the price will be update once
+  uint64 interval = 5;
+  //tokenfeeder is initialized with forever live, update the End parameters by voting, 
+  // and will off service by the end
+  // this is set by updateParams, and the EndRoundID will be update by related. excluded, 
+  // will not work if current height >=EndBlock
+  uint64 end_block = 6;
```
MaxMustermann2, line 44: I like this design

proto/exocore/oracle/tx.proto
```diff
@@ -0,0 +1,53 @@
...
+message MsgCreatePrice {
+  // creator tells which is the message sender and should sign this message
+  string creator = 1 [(cosmos_proto.scalar) = "cosmos.ValidatorAddressString"];
+  //refer to id from Params.TokenFeeders, 0 is reserved, invalid to use
+  uint64 feeder_id = 2 [(gogoproto.customname) = "FeederID"];
+
+  // prices price with its corresponding source
+  repeated PriceSource prices = 3;
+  //on which block commit does this message be built on
+  uint64 based_block = 4;
```
MaxMustermann2, line 31: Why use this instead of `roundID` ?

x/oracle/client/cli/tx_create_price.go
```diff
@@ -0,0 +1,91 @@
...
+				price := args[i]
+				timestamp := args[i+1]
+				detID := args[i+2]
+				argLength -= 3
+				i += 3
+				prices[0].Prices = append(prices[0].Prices, &types.PriceTimeDetID{
+					Price:     price,
+					Decimal:   int32(decimal),
+					Timestamp: timestamp,
+					DetID:     detID,
```
MaxMustermann2, line 67: Can you please explain when this would be supplied by the validator ?

x/oracle/keeper/aggregator/aggregator.go
```diff
@@ -0,0 +1,171 @@
...
+)
+
+type priceWithTimeAndRound struct {
+	price      *big.Int
+	decimal    int32
+	timestamp  string
+	detRoundID string // roundId from source if exists
+}
+
+type reportPrice struct {
```
MaxMustermann2, line 17: Please add a comment here describing this `struct` and the one above for easier reading. It looks like it is storing the prices reported by each validator and their vote power.

x/oracle/keeper/aggregator/aggregator.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	r.price = common.BigIntList(tmp).Median()
+	return r.price
+}
+
+type aggregator struct {
+	finalPrice *big.Int
+	reports    []*reportPrice
+	// total valiadtor power who has submitted pice
+	reportPower *big.Int
```
MaxMustermann2, line 42: Should this be an `int64` given the definition of `abci.ValidatorUpdate` ?

x/oracle/keeper/aggregator/aggregator.go
```diff
@@ -0,0 +1,171 @@
...
+						if priceTmp := reportTmp.prices[pSource.SourceID]; priceTmp != nil && priceTmp.price != nil {
+							pTR.price = new(big.Int).Set(priceTmp.price)
+							pTR.detRoundID = priceTmp.detRoundID
+							pTR.timestamp = priceTmp.timestamp
+						}
+					}
+				}
+				report.prices[pSource.SourceID] = pTR
+			}
+			// skip if this DS's slot exists, DS's value only updated by calculator
```
MaxMustermann2, line 102: should we error out here?

x/oracle/keeper/aggregator/aggregator.go
```diff
@@ -0,0 +1,171 @@
...
+
+func (agg *aggregator) aggregate() *big.Int {
+	if agg.finalPrice != nil {
+		return agg.finalPrice
+	}
+	// TODO: implemetn different MODE for definition of consensus,
+	// currently: use rule_1+MODE_1: {rule:specified source:`chainlink`, MODE: asap when power exceeds the threshold}
+	// 1. check OVA threshold
+	// 2. check IVA consensus with rule, TODO: for v1 we only implement with mode=1&rule=1
+	if common.ExceedsThreshold(agg.reportPower, agg.totalPower) {
```
MaxMustermann2, line 147: The naming of this function is a bit confusing. Consider `IsQuorumMet` or similar?

x/oracle/keeper/aggregator/aggregator.go
```diff
@@ -0,0 +1,171 @@
...
+		// TODO: this is kind of a mock way to suite V1, need update to check with params.rule
+		// check if IVA all reached consensus
```
MaxMustermann2, line 149: To confirm, this means no IVA is implemented right?

x/oracle/keeper/aggregator/calculator.go
```diff
@@ -0,0 +1,165 @@
...
+type priceAndPower struct {
+	price *big.Int
+	power *big.Int
+}
+
+// for a specific DS round, it could have multiple values provided by different validators(should not be true if there's no malicious validator)
+type roundPrices struct { // 0 means NS
+	detID     string
+	prices    []*priceAndPower
+	price     *big.Int
```
MaxMustermann2, line 27: Should we use Cosmos SDK math here instead?

x/oracle/keeper/aggregator/context.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+type roundInfo struct {
+	// this round of price will start from block basedBlock+1, the basedBlock served as a trigger to notify validators to submit prices
+	basedBlock uint64
+	// next round id of the price oracle service, price with the id will be record on block basedBlock+1 if all prices submitted by validators(for v1, validators serve as oracle nodes) get to consensus immediately
+	nextRoundID uint64
+	// indicate if this round is open for collecting prices or closed in either condition that success with a consensused price or not
+	// 1: open, 2: closed
+	status int32
```
MaxMustermann2, line 26: Convert to enum.

x/oracle/keeper/aggregator/filter.go
```diff
@@ -0,0 +1,85 @@
...
+		validatorNonce:  make(map[string]*common.Set[int32]),
+		validatorSource: make(map[string]*common.Set[string]),
+	}
+}
+
+func (f *filter) newVNSet() *common.Set[int32] {
+	return common.NewSet[int32](f.maxNonce)
+}
+
+func (f *filter) newVSSet() *common.Set[string] {
```
MaxMustermann2, line 32: Consider renaming to `newValidatorNonceSet` or even `newNonceSet` and `newValidatorSourceSet` or `newSourceSet` for clarity.

x/oracle/keeper/aggregator/worker.go
```diff
@@ -0,0 +1,62 @@
...
+type worker struct {
+	sealed  bool
+	price   string
+	decimal int32
+	// mainly used for deterministic source data to check conflicts and validation
+	f *filter
+	// used to get to consensus on deterministic source's data
+	c *calculator
+	// when enough data(exceeds threshold) collected, aggregate to conduct the final price
+	a   *aggregator
```
MaxMustermann2, line 20: The single alphabet names should be changed for clarity.

x/oracle/keeper/cache/caches.go
```diff
@@ -0,0 +1,227 @@
...
+	"github.com/ExocoreNetwork/exocore/x/oracle/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+var zeroBig = big.NewInt(0)
+
+type (
+	ItemV map[string]*big.Int
+	ItemP *common.Params
+	ItemM struct {
```
MaxMustermann2, line 16: Please rename these to something more readable.

x/oracle/keeper/common/expected_keepers.go
```diff
@@ -0,0 +1,44 @@
...
+	"cosmossdk.io/math"
+	"github.com/ExocoreNetwork/exocore/x/oracle/types"
+	abci "github.com/cometbft/cometbft/abci/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	stakingkeeper "github.com/cosmos/cosmos-sdk/x/staking/keeper"
+	stakingTypes "github.com/cosmos/cosmos-sdk/x/staking/types"
+)
+
+type KeeperOracle interface {
+	KeeperStaking
```
MaxMustermann2, line 13: Why embed this instead of creating a new interface?

x/oracle/keeper/cache/caches.go
```diff
@@ -0,0 +1,227 @@
...
+	return true
+}
+
+func (c *Cache) CommitCache(ctx sdk.Context, reset bool, k common.KeeperOracle) {
+	if len(c.msg) > 0 {
+		c.msg.commit(ctx, k)
+		c.msg = make(map[uint64][]*ItemM)
+	}
+
+	if c.validators.update {
```
MaxMustermann2, line 199: This is not set in `GetCache` so is it true in the `EndBlocker` ? Not sure if I missed something.

x/oracle/keeper/common/types.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+	}
+	return false
+}
+
+func (s *Set[T]) Length() int {
+	return s.size
+}
+
+func NewSet[T comparable](length int) *Set[T] {
```
MaxMustermann2, line 142: Nice!

x/oracle/keeper/prices.go
```diff
@@ -0,0 +1,170 @@
...
+			k.cdc.MustUnmarshal(b, val.PriceList[i])
+			found = true
+		}
+	}
+
+	return
+}
+
+// RemovePrices removes a prices from the store
+func (k Keeper) RemovePrices(
```
MaxMustermann2, line 48: Can we do some pruning of this data if it gets old ?

x/oracle/keeper/query_prices.go
```diff
@@ -0,0 +1,56 @@
...
+
+	val, found := k.GetPrices(
+		ctx,
+		req.TokenId,
+	)
+	if !found {
+		return nil, status.Error(codes.NotFound, "not found")
+	}
+
+	return &types.QueryGetPricesResponse{Prices: val}, nil
```
MaxMustermann2, line 55: paginate ?

x/oracle/keeper/recent_msg.go
```diff
@@ -0,0 +1,77 @@
+package keeper
```
MaxMustermann2, line 1: When is the data used? Can you explain briefly?

x/oracle/keeper/recent_params.go
```diff
@@ -0,0 +1,76 @@
+package keeper
```
MaxMustermann2, line 1: When is the data used? Can you explain briefly?

x/oracle/keeper/single.go
```diff
@@ -0,0 +1,146 @@
...
+	pTmp := common.Params(p)
+	agc.SetParams(&pTmp)
+	// set params cache
+	c.AddCache(cache.ItemP(&pTmp))
+
+	totalPower := big.NewInt(0)
+	validatorPowers := make(map[string]*big.Int)
+	k.IterateBondedValidatorsByPower(ctx, func(_ int64, validator stakingtypes.ValidatorI) bool {
+		power := big.NewInt(validator.GetConsensusPower(validator.GetBondedTokens()))
+		// addr := string(validator.GetOperator())
```
mikebraver: Delete unnecessary comment

x/oracle/keeper/recent_params.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"github.com/ExocoreNetwork/exocore/x/oracle/types"
+	"github.com/cosmos/cosmos-sdk/store/prefix"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// SetRecentParams set a specific recentParams in the store from its index
+func (k Keeper) SetRecentParams(ctx sdk.Context, recentParams types.RecentParams) {
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), types.KeyPrefix(types.RecentParamsKeyPrefix))
```
mikebraver, line 11: Is this only used in genesis initialize?

x/oracle/keeper/msg_server_create_price.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"context"
+	"strconv"
+
+	"github.com/ExocoreNetwork/exocore/x/oracle/types"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// CreatePrice proposes price for new round of specific tokenFeeder
+func (ms msgServer) CreatePrice(goCtx context.Context, msg *types.MsgCreatePrice) (*types.MsgCreatePriceResponse, error) {
```
mikebraver, line 12: Wouldn't this be called by validator?

x/oracle/keeper/prices.go
leonz789: Yes, we can add the pruning feature as config in further PR.

x/oracle/keeper/recent_params.go
leonz789: No, this is used to refill the cache

x/oracle/keeper/query_prices.go
leonz789: Yes, i will update in a new PR with related test

x/oracle/keeper/recent_msg.go
leonz789: This is used for refill the cache which stores price messages for aggregation:
since we have a window of more than one block to collect the prices from all validators, if this process is interrupted before the final aggregated price written to block, we would like to have some way to retrieve these messages. After we got the final price, the messages could be able to remove, so we just need the recent messages within the window.

x/oracle/keeper/recent_params.go
leonz789: This is like the recent_msgs, which is used to refill the cache:
the aggregation is affected by 3 elements:
1. validatorset(: `votingPower`)
2. params(: `tokenFeeder`)
3. messages (: including price_with_`votingPower` for `tokenFeeder`)
So when we do the refill, we just use these 3 data to reproduce the aggregation process.
But i think the recent_params could be optimized with carefully check like validator_update_block so that we don't need to keep the history params, currently this is a straightforward way. I will update in a new PR

x/oracle/keeper/msg_server_create_price.go
leonz789: This serves a a message service, and the message should be signed by validator

app/app.go
leonz789: updated

app/app.go
leonz789: updated

app/app.go
leonz789: udpated

proto/exocore/oracle/price.proto
leonz789: corrected

x/oracle/keeper/single.go
leonz789: updated

mikebraver (approved):
fix the action error then

MaxMustermann2 (approved):

PR #35, opened 2024-04-17
feat(precompile): add `getClientChains` function

MaxMustermann2:
This function can be used by the `ExocoreGateway` to fetch all currently registered client chains. This list is used to send a `markBootstrapped` transaction to all the destination chains.

TODO: at this point, the function to mark bootstrap on all chains must be triggered manually. This call should be automated in the future.

### DIFF
--- a/.github/workflows/consensuswarn.yml
+++ b/.github/workflows/consensuswarn.yml
@@ -20,7 +20,7 @@ jobs:
     steps:
       # This is used for warning when a PR touches any of the roots, or any function or method directly or indirectly called by a root
       - uses: actions/checkout@v4
-      - uses: orijtech/consensuswarn@main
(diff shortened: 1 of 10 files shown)


MaxMustermann2:
The "Consensus Warn / main (pull_request_target)" is panicking since its dependency was upgraded. I have pinned the version at a commit hash prior to the upgrade. However, the workflow runs on the base branch and the pinned version change will only be affective after this PR is merged.

bwhour (approved):

TimmyExogenous (approved):
In the future, maybe we need to provide a general precompile contract for `assets` module to enable querying all states, not only the client chains.

mikebraver (approved):
LGTM

mikebraver:
Please fix github action errors

leonz789 (approved):

PR #37, opened 2024-04-18
fix(docker): fix bugs for initialize clusters of nodes

mikebraver:
## Description
Add libstdc++ lib dependency and other fix

----

### DIFF
--- a/networks/local/exocore/Dockerfile
+++ b/networks/local/exocore/Dockerfile
@@ -10,7 +10,7 @@ RUN LEDGER_ENABLED=false make build
 
 #####################################
 FROM alpine:3.18 AS run
-RUN apk add --no-cache bash=5.2.15-r5 curl=8.5.0-r0 jq=1.6-r4 \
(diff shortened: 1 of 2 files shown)


networks/local/exocore/Dockerfile
```diff
@@ -25,7 +25,7 @@ COPY ./networks/local/exocore/wrapper.sh /usr/bin/wrapper.sh
 COPY --from=build /go/work/build/exocored /exocore
 
 # Use the created non-root user
-USER exocoreuser
+USER exocore
```
cloud8little, line 28: can we use root directly?

trestinlsd (approved):

trestinlsd (commented):
Local command make lint needs to be executed to resolve errors

bwhour (approved):

networks/local/exocore/Dockerfile
mikebraver: Better not,  will cause safety concern

MaxMustermann2 (approved):

PR #39, opened 2024-04-21
[refactor] New voting power implementation

TimmyExogenous:
# Description

This PR implements the new voting power design, please refer to the above documentation for more details.
[voting power design](https://www.notion.so/exocore/design-of-Exocore-voting-power-c2a8eaafce6e46938ba6b6b60faed9d2)

Customizing epoch for every AVS hasn't been implemented here, this might be implemented in the AVS module. So for dogfood, voting power will still be updated each block,  because the mock AVS configures one block as the epoch duration of the dogfood.
Additionally, the mock AVS configures all registered assets as the default supported assets of all AVSs.  

## change 

* fix the bug regarding querying parameters in the assets module.
[bug info](https://www.notion.so/exocore/QA-Testing-028cab07062b489f9ad373c1ce59b894?pvs=4#dfe328be5bbc4ab2b81467712514cdfc)
* using the share mechanism in the delegation module
* refactor some functions about state updating and retrieving in the assets and delegation modules. 
* implementing new voting power calculation in the operator module
* removing the voting power state storage for stakers
* implementing the instant and no-instantaneous slash in the operator module
* fix all existing tests influenced by the change
* merge the develop branch and fix conflict

# todo
* implementing the customized epoch for every AVS in the AVS module


x/delegation/keeper/delegation_state.go
```diff
@@ -11,97 +11,133 @@ import (
...
+	for ; iterator.Valid(); iterator.Next() {
+		var amounts delegationtype.DelegationAmounts
+		k.cdc.MustUnmarshal(iterator.Value(), &amounts)
+		keys, err := delegationtype.ParseStakerAssetIDAndOperatorAddrFromKey(iterator.Key())
+		if err != nil {
+			return err
+		}
+		err = opFunc(keys, &amounts)
+		if err != nil {
+			return err
```
bwhour, line 32: Here, if opFunc execution fails, an error is returned immediately and subsequent delegates are not processed

TimmyExogenous: The error should be returned to abort the iteration when calling this function, and the caller will handle the returned error.

x/operator/keeper/slash.go
```diff
@@ -15,224 +17,283 @@ import (
...
-	height := ctx.BlockHeight()
-	if slashEventHeight > height {
-		return errorsmod.Wrap(types.ErrSlashOccurredHeight, fmt.Sprintf("slashEventHeight:%d,curHeight:%d", slashEventHeight, height))
-	}
-
-	// get the state when the slash occurred
-	// get the opted-in info
-	historicalSateCtx, err := assetstype.ContextForHistoricalState(ctx, slashEventHeight)
+// Slash performs all slash events include instant slash and no-instantaneous slash.
+func (k *Keeper) Slash(ctx sdk.Context, parameter *SlashInputInfo) error {
```
bwhour, line 272: The handling of different types of slash mechanisms (instant and non-instant slash) can be refactored using the **Strategy Pattern**. The Strategy Pattern allows the implementation of an algorithm or operation to be chosen at runtime so that the algorithm changes independently of the client using it. 

1. First, define an interface (Strategy Interface) that represents the different punishment strategies, 
2. and then implement this interface for each punishment mechanism. 
3. Finally, create a Context class that will select and use the appropriate strategy for each situation.

proto/exocore/operator/v1/tx.proto
```diff
@@ -80,6 +80,18 @@ message OptedInAssetState {
...
+// SlashType represents different slash type
+enum SlashType {
+  // SLASH_TYPE__UNSPECIFIED represents the unspecified slash type
+  SLASH_TYPE_UNSPECIFIED = 0;
+  // SLASH_TYPE_INSTANT_SLASH represents the slash events that will be handled instantly after occurring
+  // so the slash will reduce the amount of related operator's assets pool.
+  SLASH_TYPE_INSTANT_SLASH = 1;
+  // NoInstantaneousSlash indicates that the slash event will be processed after a certain
+  // period of time, thus requiring a reduction in the share of the corresponding staker.
+  SLASH_TYPE_NO_INSTANTANEOUS_SLASH = 2;
```
mikebraver, line 92: Better rename to `SLASH_TYPE_NON_INSTANTANEOUS_SLASH `

x/assets/keeper/client_chain_asset.go
```diff
@@ -68,6 +69,27 @@ func (k Keeper) GetStakingAssetInfo(ctx sdk.Context, assetID string) (info *asse
...
 }
 
+func (k Keeper) GetAssetsDecimal(ctx sdk.Context, assets map[string]interface{}) (decimals map[string]uint32, err error) {
+	if assets == nil {
+		return nil, errorsmod.Wrap(assetstype.ErrInputPointerIsNil, "assets is nil")
+	}
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), assetstype.KeyPrefixReStakingAssetInfo)
+	decimals = make(map[string]uint32, 0)
+	for assetID := range assets {
+		ifExist := store.Has([]byte(assetID))
```
mikebraver: I think we can use `Get()` directly, it'll return nil if didn't exist, update to:
```
value := store.Get([]byte(assetID))
if value == nil {
    return nil, assetstype.ErrNoClientChainAssetKey
}

```

x/assets/keeper/operator_asset.go
```diff
@@ -93,24 +99,29 @@ func (k Keeper) UpdateOperatorAssetState(ctx sdk.Context, operatorAddr sdk.Addre
...
 		k.cdc.MustUnmarshal(iterator.Value(), &amounts)
 		keys, err := assetstype.ParseJoinedKey(iterator.Key())
 		if err != nil {
 			return err
 		}
-		if len(keys) == 3 {
-			err = f(keys[0], keys[1], &amounts)
-			if err != nil {
-				return err
+		if assetsFilter != nil {
```
mikebraver, line 110: Is it necessary to loop all the keys if assetFilter is not nil? We can get the key directly from assetsFilter, right?

x/operator/keeper/slash.go
mikebraver: In the new version of golang, it can use generic feature, also can apply in updateAssetValue with different type:
`func UpdateValue[T any](valueToUpdate, changeValue *T) error{} `

proto/exocore/operator/v1/tx.proto
TimmyExogenous: ok

x/delegation/keeper/abci.go
```diff
@@ -53,31 +53,18 @@ func (k *Keeper) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) []abci.Valida
...
-		delegatorAndAmount[record.OperatorAddr] = &delegationtype.DelegationAmounts{
-			WaitUndelegationAmount:  recordAmountNeg,
-			UndelegatableAfterSlash: record.ActualCompletedAmount.Neg(),
+		deltaAmount := &delegationtype.DeltaDelegationAmounts{
+			WaitUndelegationAmount: recordAmountNeg,
 		}
-		err = k.UpdateDelegationState(ctx, record.StakerID, record.AssetID, delegatorAndAmount)
+		_, err = k.UpdateDelegationState(ctx, record.StakerID, record.AssetID, record.OperatorAddr, deltaAmount)
 		if err != nil {
 			panic(err)
```
MaxMustermann2, line 64: Please remove the panic using cached context or add a TODO to remove it later.

x/delegation/keeper/cross_chain_tx_process.go
```diff
@@ -140,39 +54,40 @@ func (k *Keeper) delegateTo(
...
+	deltaAmount := &delegationtype.DeltaDelegationAmounts{
+		UndelegatableShare: share,
 	}
-	err = k.UpdateDelegationState(ctx, stakerID, assetID, delegatorAndAmount)
+	_, err = k.UpdateDelegationState(ctx, stakerID, assetID, params.OperatorAddress.String(), deltaAmount)
 	if err != nil {
 		return err
 	}
-	err = k.UpdateStakerDelegationTotalAmount(ctx, stakerID, assetID, params.OpAmount)
+	err = k.SetStakersForOperator(ctx, params.OperatorAddress.String(), assetID, stakerID)
```
MaxMustermann2: `AppendStakerForOperator` ?

x/delegation/keeper/cross_chain_tx_process.go
```diff
@@ -124,7 +38,7 @@ func (k *Keeper) delegateTo(
 
 	// update the related states
 	if params.OpAmount.IsNegative() {
```
MaxMustermann2: Should this be `!IsPositive()` to match the error ?

x/operator/types/expected_keepers.go
```diff
@@ -94,26 +129,54 @@ func (MockOracle) GetPriceChangeAssets(_ sdk.Context) (map[string]*PriceChange,
...
+		return nil, err
+	}
+	for assetID := range allAssets {
+		ret[assetID] = nil
+	}
 	return ret, nil
 }
 
-func (MockAvs) GetAvsSlashContract(_ sdk.Context, _ string) (string, error) {
+func (a MockAvs) GetAvsSlashContract(_ sdk.Context, _ string) (string, error) {
```
bwhour: can standardize the form. Capitalize all AVS.

bwhour (approved):
could merge first for integration test and then optimize

precompiles/slash/methods.go
```diff
@@ -15,7 +15,7 @@ const (
 	MethodSlash = "submitSlash"
 )
 
-// SubmitSlash Slash assets to the staker, that will change the state in slash module.
+// SubmitSlash NoInstantaneousSlash assets to the staker, that will change the state in slash module.
```
MaxMustermann2: Why this edit?

precompiles/slash/setup_test.go
```diff
@@ -26,7 +26,7 @@ func TestPrecompileTestSuite(t *testing.T) {
 
 	// Run Ginkgo integration tests
 	RegisterFailHandler(Fail)
-	RunSpecs(t, "Slash Precompile Suite")
+	RunSpecs(t, "NoInstantaneousSlash Precompile Suite")
```
MaxMustermann2: Why this edit?

proto/exocore/delegation/v1/query.proto
```diff
@@ -17,43 +17,45 @@ message DelegationInfoReq {
   string asset_id = 2 [(gogoproto.customname) = "AssetID"];
 }
 
+// StakerMap is a map to save a batch of stakers
+message StakerMap {
+  // stakers is a map, the bool value is useless, just because the protobuf
+  // doesn't support the nil value. The key is stakerID
+  map<string,bool> stakers = 1;
```
MaxMustermann2, line 24: Why not repeated string?

x/assets/types/keys.go
```diff
@@ -95,11 +95,11 @@ var (
 	KeyPrefixReStakingAssetInfo = []byte{prefixRestakingAssetInfo}
 
 	// KeyPrefixReStakerAssetInfos restakerID = clientChainAddr+'_'+ExoCoreChainIndex
-	// KeyPrefixReStakerAssetInfos key->value: restakerID+'_'+AssetID->ReStakerAssetInfo
+	// KeyPrefixReStakerAssetInfos key->value: restakerID+'/'+AssetID->ReStakerAssetInfo
 	// or restakerID->mapping(AssetID->ReStakerAssetInfo)?
 	KeyPrefixReStakerAssetInfos = []byte{prefixRestakerAssetInfo}
 
-	// KeyPrefixOperatorAssetInfos key->value: operatorAddr+'_'+AssetID->OperatorAssetInfo
+	// KeyPrefixOperatorAssetInfos key->value: operatorAddr+'/'+AssetID-> OperatorAssetInfo
```
MaxMustermann2, line 65: Can you please document anywhere where `_` or `/` are used ?

x/delegation/keeper/share.go
```diff
@@ -0,0 +1,210 @@
...
+
+// TokensFromShares calculate the token amount of provided shares, then truncated to Int
+// It uses `LegacyDec.Quo` to calculate the quotient, `LegacyDec.Quo` perform a bankers
+// rounding quotient, so the calculated token amount may be either larger or smaller
+// due to precision rounding issues. But it's acceptable, because bankers rounding balances the
+// deviation caused by precision, especially when a large number of restakers undertake
+// undelegation. Additionally, the last undelegation from an operator will undelegate all
+// remaining token to avoid the calculated token amount is bigger than the remaining token
+// caused by the bankers rounding.
+func TokensFromShares(stakerShare, totalShare sdkmath.LegacyDec, operatorAmount sdkmath.Int) (sdkmath.Int, error) {
```
MaxMustermann2: Instead of `operatorAmount`, use `totalAmount` for clarity.

x/delegation/keeper/share.go
```diff
@@ -0,0 +1,210 @@
...
+	return (stakerShare.MulInt(operatorAmount)).Quo(totalShare).TruncateInt(), nil
+}
+
+// SharesFromTokens returns the shares of a delegation given a delegated amount. It
+// returns an error if the validator has no tokens.
+// It uses `LegacyDec.QuoInt` to calculate the quotient, the calculated result will
+// be truncated through the truncation implemented by golang's standard big.Int.
+// So the calculated share might tend to be smaller, but it seems acceptable, because
+// we need to make sure the staker can't get a bigger share than they should get.
+func SharesFromTokens(totalShare sdkmath.LegacyDec, stakerAmount, operatorAmount sdkmath.Int) (sdkmath.LegacyDec, error) {
```
MaxMustermann2: Same here.

x/delegation/keeper/share.go
```diff
@@ -0,0 +1,210 @@
...
+			return share, err
+		}
+	}
+	return share, nil
+}
+
+// ValidateUndeleagtionAmount validates that a given undelegation amount is
+// valid based on upon the converted shares. If the amount is valid, the total
+// amount of respective shares is returned, otherwise an error is returned.
+func (k Keeper) ValidateUndeleagtionAmount(
```
MaxMustermann2: Typo: `ValidateUndelegationAmount`

x/delegation/keeper/share.go
```diff
@@ -0,0 +1,210 @@
...
+		return share, err
+	}
+
+	if share.GT(delegationInfo.UndelegatableShare) {
+		return share, delegationtypes.ErrInsufficientShares
+	}
+
+	// Depending on the share, amount can be smaller than unit amount(1stake).
+	// If the remain amount after unbonding is smaller than the minimum share,
+	// it's completely unbonded to avoid leaving dust shares.
```
MaxMustermann2, line 101: Well written.

x/operator/types/expected_keepers.go
```diff
@@ -94,26 +129,54 @@ func (MockOracle) GetPriceChangeAssets(_ sdk.Context) (map[string]*PriceChange,
...
+}
 
-func (MockAvs) GetAvsSupportedAssets(_ sdk.Context, _ string) (map[string]interface{}, error) {
-	// set USDT as the default asset supported by AVS
+func (a MockAvs) GetAvsSupportedAssets(ctx sdk.Context, _ string) (map[string]interface{}, error) {
+	// set all registered assets as the default asset supported by mock AVS
 	ret := make(map[string]interface{})
-	usdtAssetID := "0xdac17f958d2ee523a2206206994597c13d831ec7_0x65"
-	ret[usdtAssetID] = nil
+	allAssets, err := a.AssetsKeeper.GetAllStakingAssetsInfo(ctx)
```
MaxMustermann2, line 150: Good call.

MaxMustermann2 (commented):
I still have to review the slashing bits but others look reasonable. Good job.

x/operator/keeper/slash.go
TimmyExogenous: Seems like the generic type can't be used to reduce the repeated code because `math.LegacyDec` and `math.Int` have some customized functions, such as `Neg() (res Int)`, which can't be handled by generic types or interfaces.

TimmyExogenous: > The handling of different types of slash mechanisms (instant and non-instant slash) can be refactored using the **Strategy Pattern**. The Strategy Pattern allows the implementation of an algorithm or operation to be chosen at runtime so that the algorithm changes independently of the client using it.
> 
> 1. First, define an interface (Strategy Interface) that represents the different punishment strategies,
> 2. and then implement this interface for each punishment mechanism.
> 3. Finally, create a Context class that will select and use the appropriate strategy for each situation.

Only one slash might remain, depending on the new slash design implementation. So the change should be considered in light of the new slash implementation.

proto/exocore/operator/v1/tx.proto
TimmyExogenous: The change should be considered in light of the new slash implementation.

x/assets/keeper/client_chain_asset.go
TimmyExogenous: It will be fixed in the subsequent PR.

x/operator/types/expected_keepers.go
TimmyExogenous: Yeah, it will be fixed in the subsequent PR.

x/assets/keeper/operator_asset.go
TimmyExogenous: It depends on whether AVS or the operator has more assets. If the operator has more assets, then range assetsFilter is indeed more efficient; if AVS has more, then iterating over all assets of the operator is more efficient. And since the system has a small number of assets, both implementations are acceptable.

x/delegation/keeper/abci.go
TimmyExogenous: Okay, I will add a to-do comment in the subsequent PR.

x/delegation/keeper/cross_chain_tx_process.go
TimmyExogenous: rename it in the next PR.

x/delegation/keeper/cross_chain_tx_process.go
TimmyExogenous: Yeah, has been fixed it in the subsequent PR.

precompiles/slash/methods.go
TimmyExogenous: It's a mistake, will change it in the next PR.

precompiles/slash/setup_test.go
TimmyExogenous: same answer

proto/exocore/delegation/v1/query.proto
TimmyExogenous: The list needs to consider the repeated item, so use a map.

x/assets/types/keys.go
TimmyExogenous: will add some comments in the subsequent PR.

x/delegation/keeper/share.go
TimmyExogenous: Indeed, will rename it in the next PR.

x/delegation/keeper/share.go
TimmyExogenous: same answer.

x/delegation/keeper/share.go
TimmyExogenous: Thanks, it will be fixed in the next PR.

x/assets/keeper/operator_asset.go
```diff
@@ -10,27 +10,27 @@ import (
...
 	ret := make(map[string]*assetstype.OperatorAssetInfo, 0)
-	for ; iterator.Valid(); iterator.Next() {
-		var stateInfo assetstype.OperatorAssetInfo
-		k.cdc.MustUnmarshal(iterator.Value(), &stateInfo)
-		keyList, err := assetstype.ParseJoinedStoreKey(iterator.Key(), 2)
-		if err != nil {
-			return nil, err
+	opFunc := func(assetID string, state *assetstype.OperatorAssetInfo) error {
+		if _, ok := assetsFilter[assetID]; ok {
+			ret[assetID] = state
```
leonz789: 1. `IteratorAssetsForOperator` has checked `assetsFilter`, so opFunc may not need to do it again (maybe better just keep the filter in `opFunc`)
2. in `Itera...`, if the assetsFilter is nil, that means 'all assets', but here will skip all(and also will cause panic)

x/delegation/keeper/abci.go
```diff
@@ -27,14 +27,14 @@ func (k *Keeper) EndBlock(ctx sdk.Context, _ abci.RequestEndBlock) []abci.Valida
 				if record.CompleteBlockNumber <= uint64(ctx.BlockHeight()) {
 					panic(fmt.Sprintf("the reset completedHeight isn't in future,setHeight:%v,curHeight:%v", record.CompleteBlockNumber, ctx.BlockHeight()))
 				}
-				_, err = k.SetSingleUndelegationRecord(ctx, record)
-				if err != nil {
-					panic(err)
+				_, innerError = k.SetSingleUndelegationRecord(ctx, record)
```
leonz789, line 30: no declare ?

x/assets/keeper/operator_asset.go
TimmyExogenous: Yeah, thanks. I will fix it.

x/delegation/keeper/share.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+// SharesFromTokens returns the shares of a delegation given a delegated amount. It
+// returns an error if the validator has no tokens.
+// It uses `LegacyDec.QuoInt` to calculate the quotient, the calculated result will
+// be truncated through the truncation implemented by golang's standard big.Int.
+// So the calculated share might tend to be smaller, but it seems acceptable, because
+// we need to make sure the staker can't get a bigger share than they should get.
+func SharesFromTokens(totalShare sdkmath.LegacyDec, stakerAmount, totalAmount sdkmath.Int) (sdkmath.LegacyDec, error) {
+	if stakerAmount.GT(totalAmount) {
```
leonz789: Why `stakerAmount` greater than `totalAmount` result in an error return ? stake more amount than the validator currently hold should be allowed

x/operator/keeper/abci.go
```diff
@@ -2,116 +2,93 @@ package keeper
...
+	avsVotingPower := sdkmath.LegacyNewDec(0)
+	opFunc := func(operator string, votingPower *sdkmath.LegacyDec) error {
+		// clear the old voting power for the operator
+		*votingPower = sdkmath.LegacyNewDec(0)
+		usdValue, selfUSDValue, err := k.CalculateUSDValueForOperator(ctx, operator, assets, decimals, prices)
+		if err != nil {
+			return err
+		}
+		// check if self USD value is more than the minimum self delegation.
+		minimumSelfDelegation, err := k.avsKeeper.GetAVSMinimumSelfDelegation(ctx, avsAddr)
```
leonz789: this can be done once outside the iteration

x/operator/keeper/abci.go
```diff
@@ -2,116 +2,93 @@ package keeper
...
+		if err != nil {
+			return err
+		}
+		if selfUSDValue.GTE(minimumSelfDelegation) {
+			*votingPower = votingPower.Add(usdValue)
```
leonz789, line 79: 1. should `Add(selfUSDValue)` since `usdValue` is the total value of asset not specific operator
2. `votingPower` is not necessary, it equals to just add `avsVotingPower`

x/delegation/keeper/share.go
TimmyExogenous: Great catch, this check should be removed. I will fix it in the next commit

x/operator/keeper/abci.go
TimmyExogenous: Yeah, I will change it.

mikebraver (approved):

x/operator/keeper/abci.go
TimmyExogenous: No, the `selfUSDValue` is the value of assets owned by the operator itself, and the `usdValue` is the total value of a specific operator, including the assets delegated from the other stakers.
The `votingPower` is necessary, because the voting power of the operator will also be updated here. It's marked by a bool `isUpdate`

leonz789 (approved):

PR #43, opened 2024-04-24
[fix] change the type of clientChainLzID from uint16 to uint32 in precompile

TimmyExogenous:
## Description

The new version of the LayerZero contract uses `uint32` to represent chainID, so change the precompiled contracts to keep the same as LayerZero.

It seems like the parameter parsing needs to be refined in the future because lots of repeated codes need to be changed when the parameter type changes.


bwhour (approved):

mikebraver (approved):
LGTM

adu-web3 (approved):

adu-web3:
LGTM

PR #56, opened 2024-05-14
hot fix for testnet upgrade

leonz789:
## Description

Set hardfork height for node upgrade with new version.

----

Closes #XXX


x/oracle/keeper/cache/caches.go
```diff
@@ -132,18 +132,22 @@ func (c *cacheParams) commit(ctx sdk.Context, k common.KeeperOracle) {
 			index.Index = index.Index[i:]
 			break
 		}
-		k.RemoveRecentParams(ctx, b)
+		if block != 4287294 {
```
mikebraver, line 135: Is this hardcoded or should be set somewhere in the config?

PR #68, opened 2024-05-27
fix: errors returned by precompiled contract can not be caught in EVM

adu-web3:
## Description

We have many precompiled contracts, and when we call the functions of precompiled contracts in EVM, if precompiled contract returns an error, this error can not be caught neither by `try/catch` clause nor low level call clause.

A workaround is that we return the `false` value to indicate that the function of precompiled contract failed instead of return an error.

----

Closes #XXX


adu-web3:
target branch is wrong, I would close this to reopen one

PR #75, opened 2024-05-28
oracle:add fields for params

leonz789:
## Description

Add fields for params.

----

Closes #XXX


x/oracle/keeper/common/types.go
```diff
@@ -153,7 +153,7 @@ func NewSet[T comparable](length int) *Set[T] {
 }
 
 func ExceedsThreshold(power *big.Int, totalPower *big.Int) bool {
-	return new(big.Int).Mul(power, big.NewInt(ThresholdB)).Cmp(new(big.Int).Mul(totalPower, big.NewInt(ThresholdA))) > 0
+	return new(big.Int).Mul(power, big.NewInt(int64(ThresholdB))).Cmp(new(big.Int).Mul(totalPower, big.NewInt(int64(ThresholdA)))) > 0
```
bwhour, line 156: can you explain why do this int64 convert ? 
no loss of precision when performing large integer calculations ?

cloud8little:
Test passed with https://github.com/ExocoreNetwork/exocore/pull/75/commits/91f7a5ef2e864a3ff45e734965b47d3a4a75d843

PR75 oracle params vs. Legacy oracle params

```
params:
  chains:
  - desc: '-'
    name: '-'
  - desc: '-'
    name: Ethereum
  max_det_id: 5
  max_nonce: 3
  mode: 1
  rules:
  - nom: null
    source_ids: []
  - nom: null
    source_ids:
    - "0"
  sources:
  - deterministic: false
    entry: null
    name: 0 position is reserved
    valid: false
  - deterministic: true
    entry:
      offchain:
        "0": ""
      onchain: {}
    name: Chainlink
    valid: true
  threshold_a: 2
  threshold_b: 3
  token_feeders:
  - end_block: "0"
    interval: "0"
    rule_id: "0"
    start_base_block: "0"
    start_round_id: "0"
    token_id: "0"
  - end_block: "0"
    interval: "10"
    rule_id: "1"
    start_base_block: "100000000"
    start_round_id: "1"
    token_id: "1"
  tokens:
  - active: false
    asset_id: ""
    chain_id: "0"
    contract_address: ""
    decimal: 0
    name: ""
  - active: true
    asset_id: ""
    chain_id: "1"
    contract_address: 0x
    decimal: 18
    name: ETH
```

```
params:
  chains: []
  rules: []
  sources: []
  token_feeders: []
  tokens: []
```

cloud8little (approved):

bwhour (approved):

x/oracle/keeper/common/types.go
MaxMustermann2: If it is from `int32` to `int64` it is safe right?

x/oracle/keeper/single.go
```diff
@@ -181,3 +183,11 @@ func ResetCache() {
 func ResetAggregatorContextCheckTx() {
 	agcCheckTx = nil
 }
+
+func setCommonParams(p types.Params) {
```
MaxMustermann2, line 187: I do not like this approach. Why not use the database and query from it each time? Too expensive?

And is it called at all the right places it should be called in?

proto/exocore/oracle/info.proto
```diff
@@ -17,14 +17,16 @@ message Chain{
...
-  //if any, like erc20 tokens
+  // if any, like erc20 tokens
   string contract_address = 3;
   // decimal of token price
   int32 decimal = 4;
-  //set false when we stop official price oracle service for a specified token
+  // set false when we stop official price oracle service for a specified token
   bool active = 5;
+  // refer to assetID from assets module if exists
+  string asset_id = 6 [(gogoproto.customname) = "AssetID"];
```
MaxMustermann2, line 29: This is unused as of now, correct?

x/oracle/keeper/common/types.go
leonz789: big.NewInt need int64 as input, and the convert is safe

proto/exocore/oracle/info.proto
leonz789: Yes, operator use assetID to query price. Set this field to map from assetID to tokenID to keep oracle module independently.

x/oracle/keeper/single.go
leonz789: These params only used when transactions are executed, and will not be changed frequently, so keep them in memory may be better.
They are initialized with AggregatorContext, to make sure they are initialized before any transaction be executed.
These variables can be moved to AggregatorContext, will do i when refactor the code

proto/exocore/oracle/info.proto
MaxMustermann2: Why use two different identifiers ? Do you intend on merging them later so that everything is following `assetID` ?

MaxMustermann2 (approved):

proto/exocore/oracle/info.proto
leonz789: Oracle module need to keep tokenInfos for history, since tokenFeeders refer to tokenID. And assets module currently seems able to modify assetsInfo. It's possible to merge these two, but need to do some adjustments with assets module

PR #77, opened 2024-05-29
feat(oracle): add support for get price indexed by assetID

leonz789:
## Description

Add method to get prices from AssetIDs. And this PR includes some code refactor which remove the unnecessary alias struct `common.Params` for `types.Params`
## Changes
Notable changes
- feat: add method to get prices from AssetIDs by set AssetID as a filed for Token in Oracle params
- refactor: remove alis `common.Params`, and move the corresponding method to `Params`

----

Closes #XXX

## Summary by CodeRabbit

- **New Features**
  - Introduced new Oracle functionality, including additional methods for managing parameters and retrieving prices for multiple assets.

- **Bug Fixes**
  - Resolved issues related to handling of empty or invalid price rounds in the Oracle module.

- **Refactor**
  - Updated internal parameter management to improve consistency and reliability.
  - Replaced `MockOracle` with `OracleKeeper` for better Oracle functionality initialization.

- **Tests**
  - Added new test cases for multi-asset price retrieval to ensure robust Oracle operations.


MaxMustermann2:
Can you please add some information about this PR and how it's working? I see that you're looping through one of the identifiers to find the other (assetID and tokenID) whenever a query is received. Doing so may be too costly to do it as frequently as we need.

TimmyExogenous (commented):
Maybe some states can be stored in the module through `kvstore`  instead of in the `params`, such as the `tokenID`. Using `assetID` as the key and the `tokenID` as the value can remove the repeated information stored in the `Oracle` module and reduce the cost when querying the `tokenID` from the `assetID`. 
Additionally, would it be better to allocate a number as the `tokenID` for all assets in the assets module? This way, the `Oracle` module can use these states directly instead of storing them.

leonz789:
> Can you please add some information about this PR and how it's working? I see that you're looping through one of the identifiers to find the other (assetID and tokenID) whenever a query is received. Doing so may be too costly to do it as frequently as we need.

The values all cached in memory, and only updated when the params changed(through gov module which is not frequently), does not rely on IO access

leonz789:
> Maybe some states can be stored in the module through `kvstore` instead of in the `params`, such as the `tokenID`. Using `assetID` as the key and the `tokenID` as the value can remove the repeated information stored in the `Oracle` module and reduce the cost when querying the `tokenID` from the `assetID`. Additionally, would it be better to allocate a number as the `tokenID` for all assets in the assets module? This way, the `Oracle` module can use these states directly instead of storing them.

The update of this state should be controlled by gov module through proposal/vote, so just put them in params. And for the frequently query, they basically accessed from memory cache

MaxMustermann2:
> > Can you please add some information about this PR and how it's working? I see that you're looping through one of the identifiers to find the other (assetID and tokenID) whenever a query is received. Doing so may be too costly to do it as frequently as we need.
> 
> The values all cached in memory, and only updated when the params changed(through gov module which is not frequently), does not rely on IO access

I see, that seems fair. But can you please provide a small doc (2 sentences or so) about how this is working? It is not clear from the code.

@coderabbitai review

leonz789:
> > > Can you please add some information about this PR and how it's working? I see that you're looping through one of the identifiers to find the other (assetID and tokenID) whenever a query is received. Doing so may be too costly to do it as frequently as we need.
> > 
> > 
> > The values all cached in memory, and only updated when the params changed(through gov module which is not frequently), does not rely on IO access
> 
> I see, that seems fair. But can you please provide a small doc (2 sentences or so) about how this is working? It is not clear from the code.
> 
> @coderabbitai review

Updated. the unclear may be introduced by the code refactor.

x/oracle/keeper/aggregator/context.go
```diff
@@ -288,6 +288,20 @@ func (agc *AggregatorContext) GetValidatorPowers() (vp map[string]*big.Int) {
 	return agc.validatorsPower
 }
 
+func (agc *AggregatorContext) GetTokenIDFromAssetID(assetID string) int {
+	agc.params.GetTokenIDFromAssetID(assetID)
```
MaxMustermann2: Why is this called if the return value is not used?

MaxMustermann2 (approved):
This PR has more changes than the subject indicates. All look good to me.

x/oracle/keeper/aggregator/context.go
leonz789: Yes this value should be returned, updated

TimmyExogenous (approved):

leonz789:
replace mock oracle with oracle keeper in operator keeper, and set default price to 1 when tokenFeeder has not reach consensus on 1st round.

mikebraver (approved):

PR #82, opened 2024-05-30
fix(assets): require non-zero address length

MaxMustermann2:
This PR adds a check in the `ValidateGenesis` function of `x/assets` to ensure that an address length is provided at launch. It adds an associated test, and sets the address length in `local_node.sh`.

### DIFF
--- a/x/assets/types/genesis.go
+++ b/x/assets/types/genesis.go
@@ -42,6 +42,7 @@ func (gs GenesisState) Validate() error {
 				i,
 			)
 		}
+		// this is our primary method of cross-chain communication.
 		if info.LayerZeroChainID == 0 {
 			return errorsmod.Wrapf(
 				ErrInvalidGenesisData,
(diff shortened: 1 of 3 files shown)


leonz789 (approved):

TimmyExogenous (approved):

x/assets/types/genesis.go
```diff
@@ -42,13 +42,23 @@ func (gs GenesisState) Validate() error {
 				i,
 			)
 		}
+		// this is our primary method of cross-chain communication.
 		if info.LayerZeroChainID == 0 {
```
bwhour, line 46: can we consider setting the default values for **LayerZeroChainID** and **AddressLength** to avoid the 0 case.

bwhour (approved):
it looks good.

cloud8little (approved):

x/assets/types/genesis.go
MaxMustermann2: As discussed, this is the client chain's address length and each chain may have different length for their addresses. It should be supplied accurately in the genesis file.

PR #94, opened 2024-06-10
fix:limit count of validators to maxValidators params when udpate

leonz789:
## Description

<p>Currently when update validators, it will allow at most `maxValidators`+1 validators</p>
----

Closes #XXX

## Summary by CodeRabbit

- **Bug Fixes**
	- Corrected a comparison condition to improve the performance and reliability of the `EndBlock` method.


MaxMustermann2 (approved):
Good catch

cloud8little (approved):

mikebraver (approved):

Issue #113, opened 2024-06-23
Record some trivial issues identified by the teammates during the code review or in the daily development process.


TimmyExogenous:
## Description

This issue records some trivial issues regarding the code style, bad naming, and trivial implementation approvement. Then we can centrally manage and address these issues. A new issue might be created to track the work when we address one of the following issues.

### issues

- [ ] In the command line, a stateless check should be performed on the input parameters, such as the operator and AVS addresses. This check can refer to the validation of the genesis state. There is a related code comment that can be referred to for this issue.
https://github.com/ExocoreNetwork/exocore/pull/61#discussion_r1647757393

- [ ] Some RPCs use a map as the response when defined in the protobuf. It appears that maps cannot support pagination. It also introduces inconsistencies when providing related functions in the precompiled contract. The relevant code is as follows:
https://github.com/ExocoreNetwork/exocore/blob/dd27216e030a40a521cf800bd5a606f9ea287e5b/proto/exocore/assets/v1/query.proto#L25

- [ ] The implementation of some RPCs and command lines should consider pagination. This issue might depend on resolving the previous issue. The relevant code is as follows:
https://github.com/ExocoreNetwork/exocore/blob/dd27216e030a40a521cf800bd5a606f9ea287e5b/proto/exocore/assets/v1/query.proto#L39

- [ ] The `layer_zero_chain_id` should be renamed to `client_chain_id` to support the client chains that do not use the layer zero protocol, such as Bitcoin.  Additionally the redundant `chain_id` should be removed for the `ClientChainInfo`. And  `ExocoreLzAppAddress` should be renamed to `ExocoreGateWayAddress`.
When addressing this issue, consider the compatibility of the launching node from the genesis, whether from the bootstrap or the general exported file.
https://github.com/ExocoreNetwork/exocore/blob/dd27216e030a40a521cf800bd5a606f9ea287e5b/proto/exocore/assets/v1/tx.proto#L24C9-L24C24

- [ ] Consider using an enum type to represent the signature type in the client chain information.
https://github.com/ExocoreNetwork/exocore/pull/116#discussion_r1658371444

- [ ] using a bool type to replace `active_usd_value`.
https://github.com/ExocoreNetwork/exocore/pull/129#discussion_r1694819583

- [ ] add a test for jail and unjail.
https://github.com/ExocoreNetwork/exocore/pull/129#pullrequestreview-2203799134

- [ ] add tests for the functions added for genesis exporting
[...]

PR #116, opened 2024-06-24
[feat] Register chains and tokens through precompile

### LINKED ISSUE #115: Non-determinism in GetClientChains function due to map iteration
The  function in  retrieves all client chain IDs but iterates over a map to do so, which could introduce non-determinism in the output order of chain IDs. This issue needs to be addressed to ensure consistent behavior, especially if the order of IDs is significant. This will be handled in a future PR.

TimmyExogenous:
## Description

Closes #115

This PR implements tokens and client chain registration functions through the precompiled contract. 
Additionally, it fix issue#115 regarding the non-determinism caused by using a map.

## Summary by CodeRabbit

- **New Features**
  - Added settings for `exocore_lz_app_address` and `meta_info` under assets, and included a new token "Tether USD."

- **Improvements**
  - Enhanced parameter clarity in deposit, withdraw, client chain, and token functions.
  - Introduced pagination for querying all client chain info.
  - Improved error messages for better clarity.
  
- **Bug Fixes**
  - Fixed naming consistency from `CheckExocoreLzAppAddr` to `CheckExocoreGatewayAddr`.
  - Corrected error messages for address length validation.

- **Refactor**
  - Renamed methods for better clarity and consistency.
  - Updated function names to reflect improved functionality.


local_node.sh
```diff
@@ -85,6 +85,7 @@ if [[ $overwrite == "y" || $overwrite == "Y" ]]; then
 	jq '.consensus_params["block"]["max_gas"]="10000000"' "$GENESIS" >"$TMP_GENESIS" && mv "$TMP_GENESIS" "$GENESIS"
 
 	# x/assets
+	jq '.app_state["assets"]["params"]["exocore_lz_app_address"]="0x3e108c058e8066da635321dc3018294ca82ddedf"' "$GENESIS" >"$TMP_GENESIS" && mv "$TMP_GENESIS" "$GENESIS"
```
MaxMustermann2, line 89: Is this the CREATE2 address?

precompiles/assets/types.go
```diff
@@ -32,27 +36,121 @@ func (p Precompile) GetDepositWithdrawParamsFromInputs(ctx sdk.Context, args []i
...
+	clientChain.AddressLength = addressLength
+
+	name, ok := args[2].(string)
+	if !ok || name == "" {
+		return nil, xerrors.Errorf(exocmn.ErrContractInputParaOrType, 2, reflect.TypeOf(args[2]), args[2])
+	}
+	clientChain.Name = name
+
+	metaInfo, ok := args[3].(string)
+	if !ok {
```
MaxMustermann2, line 93: Validate length of metaInfo non-zero ? We can look at setting a maximum size for the lengths later

precompiles/assets/types.go
```diff
@@ -32,27 +36,121 @@ func (p Precompile) GetDepositWithdrawParamsFromInputs(ctx sdk.Context, args []i
...
+		return nil, xerrors.Errorf(exocmn.ErrContractInputParaOrType, 3, reflect.TypeOf(args[2]), args[2])
+	}
+	clientChain.MetaInfo = metaInfo
+
+	signatureType, ok := args[4].(string)
+	if !ok {
```
MaxMustermann2, line 102: Same here, should we validate non-empty?

precompiles/assets/types.go
```diff
@@ -32,27 +36,121 @@ func (p Precompile) GetDepositWithdrawParamsFromInputs(ctx sdk.Context, args []i
...
+		return nil, fmt.Errorf(exocmn.ErrInvalidInputList, len(tvlLimitList), assetNumber)
+	}
+
+	for i := 0; i < assetNumber; i++ {
+		if len(assetAddrList[i]) != types.GeneralAssetsAddrLength {
+			return nil, xerrors.Errorf(exocmn.ErrInvalidAddrLength, len(assetAddrList[i]), types.GeneralClientChainAddrLength)
+		}
+		assetAddr := assetAddrList[i][:clientChainAddrLength]
+		assets = append(assets, types.AssetInfo{
+			Address:          hexutil.Encode(assetAddr),
```
MaxMustermann2: Add TODO for non-EVM chains

x/assets/keeper/client_chain.go
```diff
@@ -35,16 +35,52 @@ func (k Keeper) GetClientChainInfoByIndex(ctx sdk.Context, index uint64) (info *
...
+// IterateAllClientChains iterates all client chains, and the `opFunc` will be called for
+// each client chain. As for the `isUpdate`, it a flag to indicate if the client chain
+// info handled by the `opFunc` will be restored.
+func (k Keeper) IterateAllClientChains(ctx sdk.Context, isUpdate bool, opFunc func(clientChain *assetstype.ClientChainInfo) error) error {
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), assetstype.KeyPrefixClientChainInfo)
+	iterator := sdk.KVStorePrefixIterator(store, nil)
 	defer iterator.Close()
-
-	ret := make(map[uint64]*assetstype.ClientChainInfo, 0)
 	for ; iterator.Valid(); iterator.Next() {
```
MaxMustermann2, line 45: For my understanding, this is the part that makes it deterministic right? The iteration order is sorted in the store.

MaxMustermann2 (approved):
Some questions, but otherwise looks good.

precompiles/assets/IAssets.sol
```diff
@@ -4,46 +4,75 @@ pragma solidity >=0.8.17;
...
     function getClientChains() external view returns (bool, uint32[] memory);
-}
 
+    /// TRANSACTIONS
+    /// @dev register some client chain to allow token registration from that chain, staking
+    /// from that chain, and other operations from that chain.
+    /// @param clientChainID is the layerZero chainID if it is supported.
+    //  It might be allocated by Exocore when the client chain isn't supported
+    //  by layerZero
+    function registerClientChain(
```
MaxMustermann2, line 54: These should be added to `IsTransaction`

precompiles/reward/parser.go
```diff
@@ -47,7 +47,7 @@
 		return nil, fmt.Errorf(exocmn.ErrContractInputParaOrType, 2, reflect.TypeOf(args[0]), stakerAddr)
 	}
 	if len(assetAddr) != types.GeneralClientChainAddrLength {
-		return nil, fmt.Errorf(exocmn.ErrInputClientChainAddrLength, len(assetAddr), types.GeneralClientChainAddrLength)
+		return nil, fmt.Errorf(exocmn.ErrInvalidAddrLength, len(assetAddr), types.GeneralClientChainAddrLength)
```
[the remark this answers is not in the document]
MaxMustermann2: `stakerAddr` not `WithdrawRewardAddress`

MaxMustermann2 (approved):

local_node.sh
TimmyExogenous: No, it's the default address from the above local mnemonic. It is used to start the local node, then directly use this EOA address to test all precompiled contract interfaces, without relying on the client chain and Exocore gateway contract.

x/assets/keeper/client_chain.go
TimmyExogenous: Yeah

precompiles/assets/IAssets.sol
TimmyExogenous: Yeah, thanks

precompiles/assets/types.go
TimmyExogenous: It appears that the compatibility of non-EVM chains will be addressed in the Exocore gateway contract. However, it's still pending, so I will add a code comment.

precompiles/assets/types.go
TimmyExogenous: The initial version had a validation for this. I removed it to keep it consistent because it isn't validated in the genesis validation and this field had been set in local_node.sh. It's better to set a maximum size. I can add a maximum and the validation in both the precompile and the genesis validation.

precompiles/assets/types.go
TimmyExogenous: This field can be left empty, and there will be a default signature type when the input is empty. However, the final decision will be made after researching the signature algorithms of most blockchains.

precompiles/assets/types.go
```diff
@@ -1,7 +1,11 @@
 package assets
 
 import (
+	"fmt"
 	"math/big"
+	"reflect"
```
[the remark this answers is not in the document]
TimmyExogenous: Thanks, I will remove the usage of reflect.

precompiles/assets/assets_test.go
```diff
@@ -59,7 +59,7 @@ func paddingClientChainAddress(input []byte, outputLength int) []byte {
 	return input
 }
 
-// TestRunDepositTo tests DepositAndWithdraw method through calling Run function..
+// TestRunDepositTo tests DepositOrWithdraw method through calling Run function..
```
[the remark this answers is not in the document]
TimmyExogenous: Maybe just renaming `TestRunWithdrawThroughClientChain` to `TestRunWithdrawPrincipal` is fine.

precompiles/assets/types.go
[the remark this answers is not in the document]
TimmyExogenous: I will avoid the usage of reflect

x/assets/keeper/params.go
```diff
@@ -39,22 +39,22 @@ func (k Keeper) GetParams(ctx sdk.Context) (*assetstypes.Params, error) {
 	return ret, nil
 }
 
-func (k Keeper) GetExocoreLzAppAddress(ctx sdk.Context) (common.Address, error) {
+func (k Keeper) GetExocoreGatewayAddress(ctx sdk.Context) (common.Address, error) {
```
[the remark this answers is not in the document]
TimmyExogenous: Seem like the error has been handled in `GetExocoreGatewayAddress`

precompiles/common/error.go
```diff
@@ -4,7 +4,9 @@ const (
 	ErrContractInputParaOrType = "the contract input parameter type or value error,arg index:%d, type is:%s,value:%v"
 	ErrContractCaller          = "the caller doesn't have the permission to call this function"
 
-	ErrInputClientChainAddrLength = "the length of input client chain addr doesn't match,input:%d,need:%d"
+	ErrInvalidAddrLength = "the length of input client chain or asset addr doesn't match,input:%d,expected:%v"
```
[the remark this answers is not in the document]
TimmyExogenous: I just want to use an Error to address the different cases of invalid address length.

precompiles/slash/parser.go
```diff
@@ -37,7 +37,7 @@ func (p Precompile) GetSlashParamsFromInputs(ctx sdk.Context, args []interface{}
 		return nil, fmt.Errorf(exocmn.ErrContractInputParaOrType, 1, reflect.TypeOf(args[0]), assetAddr)
 	}
 	if len(assetAddr) != types.GeneralClientChainAddrLength {
-		return nil, fmt.Errorf(exocmn.ErrInputClientChainAddrLength, len(assetAddr), types.GeneralClientChainAddrLength)
+		return nil, fmt.Errorf(exocmn.ErrInvalidAddrLength, len(assetAddr), types.GeneralClientChainAddrLength)
```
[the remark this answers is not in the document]
TimmyExogenous: I just want to use an Error to address the different cases of invalid address length."

precompiles/assets/assets.go
```diff
@@ -89,14 +89,27 @@ func (p Precompile) Run(evm *vm.EVM, contract *vm.Contract, readOnly bool) (bz [
...
+		bz, err = p.RegisterClientChain(ctx, contract, method, args)
+		if err != nil {
+			ctx.Logger().Error("call assets precompile error", "module", "assets precompile", "method", method.Name, "err", err)
+			return method.Outputs.Pack(false)
+		}
+	case MethodRegisterTokens:
+		bz, err = p.RegisterTokens(ctx, contract, method, args)
+		if err != nil {
+			ctx.Logger().Error("call assets precompile error", "module", "assets precompile", "method", method.Name, "err", err)
+			return method.Outputs.Pack(false)
```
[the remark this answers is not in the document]
TimmyExogenous: Thanks, but the returned should be `method.Outputs.Pack(false)`

precompiles/slash/parser.go
```diff
@@ -46,7 +46,7 @@
 		return nil, fmt.Errorf(exocmn.ErrContractInputParaOrType, 2, reflect.TypeOf(args[0]), stakerAddr)
 	}
 	if len(assetAddr) != types.GeneralClientChainAddrLength {
-		return nil, fmt.Errorf(exocmn.ErrInputClientChainAddrLength, len(assetAddr), types.GeneralClientChainAddrLength)
+		return nil, fmt.Errorf(exocmn.ErrInvalidAddrLength, len(assetAddr), types.GeneralClientChainAddrLength)
```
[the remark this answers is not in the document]
TimmyExogenous: The codes regarding the slash and reward module should be deprecated or re-implemented.

precompiles/reward/parser.go
```diff
@@ -38,7 +38,7 @@ func (p Precompile) GetRewardParamsFromInputs(ctx sdk.Context, args []interface{
 		return nil, fmt.Errorf(exocmn.ErrContractInputParaOrType, 1, reflect.TypeOf(args[0]), assetAddr)
 	}
 	if len(assetAddr) != types.GeneralClientChainAddrLength {
-		return nil, fmt.Errorf(exocmn.ErrInputClientChainAddrLength, len(assetAddr), types.GeneralClientChainAddrLength)
+		return nil, fmt.Errorf(exocmn.ErrInvalidAddrLength, len(assetAddr), types.GeneralClientChainAddrLength)
```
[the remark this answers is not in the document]
TimmyExogenous: The codes regarding the slash and reward module should be deprecated or re-implemented.

precompiles/reward/parser.go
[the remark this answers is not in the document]
TimmyExogenous: The codes regarding the slash and reward module should be deprecated or re-implemented.

precompiles/delegation/delegation.go
```diff
@@ -120,7 +120,7 @@ func (p Precompile) Run(evm *vm.EVM, contract *vm.Contract, readOnly bool) (bz [
 // IsTransaction checks if the given methodID corresponds to a transaction or query.
 //
 // Available deposit transactions are:
-//   - DepositAndWithdraw
+//   - DepositOrWithdraw
```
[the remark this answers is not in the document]
TimmyExogenous: Thanks

x/assets/keeper/grpc_query.go
```diff
@@ -29,13 +32,26 @@
...
+		return nil
+	})
 	if err != nil {
 		return nil, err
 	}
-	return &assetstype.QueryAllClientChainInfoResponse{AllClientChainInfos: allInfo}, nil
+	return &assetstype.QueryAllClientChainInfoResponse{
+		AllClientChainInfos: res,
+		Pagination:          pageRes,
+	}, nil
```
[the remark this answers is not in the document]
TimmyExogenous: thanks

precompiles/assets/types.go
TimmyExogenous: Additionally, generate.js might need to be changed slightly after this PR is merged.

cloud8little:
Test passed. 

1. Register client chain through precompile contract.
```
exocored q assets QueAllClientChainInfo
all_client_chain_infos:
- address_length: 20
  chain_id: "0"
  exocore_chain_index: "0"
  finalization_blocks: "0"
  layer_zero_chain_id: "101"
  meta_info: Example EVM chain meta info
  name: Example EVM chain
  signature_type: ""
- address_length: 20
  chain_id: "0"
  exocore_chain_index: "0"
  finalization_blocks: "0"
  layer_zero_chain_id: "40159"
  meta_info: BSC
  name: Binance Smart Chain
  signature_type: ECDSA
pagination:
  next_key: null
  total: "2"
```

2. Register new asset through precompile

```
exocored q assets QueAllStakingAssetsInfo
all_staking_assets_info:
  0x1e867667ef16111047c2c7f6adf4612bdf80064d_0x9cdf:
    asset_basic_info:
      address: 0x1e867667ef16111047c2c7f6adf4612bdf80064d
      decimals: 18
      exocore_chain_index: "0"
      layer_zero_chain_id: "40159"
      meta_info: ""
      name: ""
      symbol: ""
      total_supply: "2000000000000000000"
    staking_total_amount: "0"
```

cloud8little (approved):

x/assets/keeper/params.go
```diff
@@ -39,22 +39,22 @@ func (k Keeper) GetParams(ctx sdk.Context) (*assetstypes.Params, error) {
...
 
-func (k Keeper) GetExocoreLzAppAddress(ctx sdk.Context) (common.Address, error) {
-	depositModuleParam, err := k.GetParams(ctx)
+func (k Keeper) GetExocoreGatewayAddress(ctx sdk.Context) (common.Address, error) {
+	param, err := k.GetParams(ctx)
 	if err != nil {
 		return common.Address{}, err
 	}
-	return common.HexToAddress(depositModuleParam.ExocoreLzAppAddress), nil
+	return common.HexToAddress(param.ExocoreLzAppAddress), nil
```
bwhour, line 47: can we rename this **ExocoreLzAppAddress** to ExocoreGatewayAddress which is easy to understand? 

Additionally, If in the future **exocoreBtcGateway** and **exocoreSolGateway**  etc. cannot be integrated with **exocoregateway**, how should this part of the logic be handled?

local_node.sh
MaxMustermann2: Ah okay - it is the `local_funded_account`. Can you please add a comment for future reference?

precompiles/assets/types.go
MaxMustermann2: > It's better to set a maximum size. I can add a maximum and the validation in both the precompile and the genesis validation.

Can you create an issue for tracking this?

local_node.sh
TimmyExogenous: Yeah

x/assets/keeper/params.go
TimmyExogenous: Yeah, it should be renamed to ExocoreGatewayAddress. However, the renaming might cause inconsistency in the data within the KVStore, so we need to address these issues together at an appropriate time. I have created an issue to document these concerns
https://github.com/ExocoreNetwork/exocore/issues/113

TimmyExogenous: > Additionally, If in the future exocoreBtcGateway and exocoreSolGateway etc. cannot be integrated with exocoregateway, how should this part of the logic be handled?

This depends on the design for Bitcoin and Solana. If we use a unified gateway address to address all client chains, then it doesn't need to be changed. We might store a list for the gateway address if we use multiple gateway contracts.

precompiles/assets/types.go
```diff
@@ -32,27 +35,132 @@ func (p Precompile) GetDepositWithdrawParamsFromInputs(ctx sdk.Context, args []i
...
+	}
+	clientChain.MetaInfo = metaInfo
+
+	signatureType, ok := args[4].(string)
+	if !ok {
+		return nil, xerrors.Errorf(exocmn.ErrContractInputParaOrType, 4, "string", args[4])
+	}
+	clientChain.SignatureType = signatureType
+
+	return &clientChain, nil
```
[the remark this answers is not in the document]
TimmyExogenous: Yeah, this error has provided clear and actionable information to the caller.

precompiles/assets/types.go
```diff
@@ -32,27 +35,132 @@
...
+	metaInfo, ok := args[5].(string)
+	if !ok {
+		return types.AssetInfo{}, fmt.Errorf(exocmn.ErrContractInputParaOrType, 5, "string", args[5])
+	}
+	if metaInfo == "" || len(metaInfo) > types.MaxChainTokenMetaInfoLength {
+		return types.AssetInfo{}, xerrors.Errorf(exocmn.ErrInvalidMetaInfoLength, metaInfo, len(metaInfo), types.MaxChainTokenMetaInfoLength)
+	}
+	asset.MetaInfo = metaInfo
+
+	return asset, nil
```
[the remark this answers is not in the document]
TimmyExogenous: It doesn't seem necessary to make such modifications.

precompiles/assets/types.go
TimmyExogenous: I have added the maximum and validation in the last commit.

precompiles/assets/tx.go
```diff
@@ -83,19 +88,63 @@ func (p Precompile) GetClientChains(
...
+		return nil, err
+	}
+	err = p.assetsKeeper.SetClientChainInfo(ctx, clientChainInfo)
+	if err != nil {
+		return nil, err
+	}
+	return method.Outputs.Pack(true)
+}
+
+func (p Precompile) RegisterTokens(
```
adu-web3:
```suggestion
func (p Precompile) RegisterToken(
```

precompiles/assets/types.go
adu-web3: I wonder if we could allow this `metaInfo` being empty? since it is extra data that is used for arbitrary data like an url or plain text. I don't check this field to be non-empty in the contract implementation.

precompiles/assets/types.go
```diff
@@ -32,27 +35,132 @@ func (p Precompile) GetDepositWithdrawParamsFromInputs(ctx sdk.Context, args []i
...
+	metaInfo, ok := args[3].(string)
+	if !ok {
+		return nil, xerrors.Errorf(exocmn.ErrContractInputParaOrType, 3, "string", args[2])
+	}
+	if metaInfo == "" || len(metaInfo) > types.MaxChainTokenMetaInfoLength {
+		return nil, xerrors.Errorf(exocmn.ErrInvalidMetaInfoLength, metaInfo, len(metaInfo), types.MaxChainTokenMetaInfoLength)
+	}
+	clientChain.MetaInfo = metaInfo
+
+	signatureType, ok := args[4].(string)
```
adu-web3, line 101: should we use an enum value instead of string if we could exhaust existed common signature types. Of course we could also leave it for future PRs if we haven't researched into this

precompiles/assets/types.go
```diff
@@ -32,27 +35,132 @@ func (p Precompile) GetDepositWithdrawParamsFromInputs(ctx sdk.Context, args []i
...
+	if len(args) != 6 {
+		return types.AssetInfo{}, xerrors.Errorf(cmn.ErrInvalidNumberOfArgs, 6, len(args))
+	}
+	asset := types.AssetInfo{}
+	clientChainID, ok := args[0].(uint32)
+	if !ok {
+		return types.AssetInfo{}, fmt.Errorf(exocmn.ErrContractInputParaOrType, 0, "uint32", args[0])
+	}
+	asset.LayerZeroChainID = uint64(clientChainID)
+	info, err := p.assetsKeeper.GetClientChainInfoByIndex(ctx, asset.LayerZeroChainID)
```
adu-web3, line 120: I see `p.assetsKeeper.GetClientChainInfoByIndex` would return error if client chain is not registered before, just comment to get your confirmation about this if my understanding is ture

precompiles/assets/types.go
```diff
@@ -32,27 +35,132 @@ func (p Precompile) GetDepositWithdrawParamsFromInputs(ctx sdk.Context, args []i
...
+	}
+	asset.Address = hexutil.Encode(assetAddr[:clientChainAddrLength])
+
+	decimal, ok := args[2].(uint8)
+	if !ok {
+		return types.AssetInfo{}, fmt.Errorf(exocmn.ErrContractInputParaOrType, 2, "uint8", args[2])
+	}
+	asset.Decimals = uint32(decimal)
+
+	tvlLimit, ok := args[3].(*big.Int)
```
adu-web3, line 141: In `ExocoreGateway`, I check that tvlLimit is not zero because zero tvl limit seems not make sense. what do you think?

precompiles/assets/types.go
```diff
@@ -32,27 +35,132 @@ func (p Precompile) GetDepositWithdrawParamsFromInputs(ctx sdk.Context, args []i
...
+		return types.AssetInfo{}, fmt.Errorf(exocmn.ErrContractInputParaOrType, 0, "uint32", args[0])
+	}
+	asset.LayerZeroChainID = uint64(clientChainID)
+	info, err := p.assetsKeeper.GetClientChainInfoByIndex(ctx, asset.LayerZeroChainID)
+	if err != nil {
+		return types.AssetInfo{}, err
+	}
+	clientChainAddrLength := info.AddressLength
+
+	assetAddr, ok := args[1].([]byte)
```
adu-web3, line 126: in`ExocoreGateway` I check that the asset address is non-zero, do you think it is needed here?

precompiles/assets/types.go
```diff
@@ -32,27 +35,132 @@ func (p Precompile) GetDepositWithdrawParamsFromInputs(ctx sdk.Context, args []i
...
+	if name == "" || len(name) > types.MaxChainTokenNameLength {
+		return types.AssetInfo{}, xerrors.Errorf(exocmn.ErrInvalidNameLength, name, len(name), types.MaxChainTokenNameLength)
+	}
+	asset.Name = name
+
+	metaInfo, ok := args[5].(string)
+	if !ok {
+		return types.AssetInfo{}, fmt.Errorf(exocmn.ErrContractInputParaOrType, 5, "string", args[5])
+	}
+	if metaInfo == "" || len(metaInfo) > types.MaxChainTokenMetaInfoLength {
```
adu-web3, line 160: same as above comment it seems reasonable to allow metaInfo being empty

adu-web3 (approved):
overall look good to me, I have some comments mainly about parameters validation for precompile functions `registerClientChain` and `registerToken`

precompiles/assets/types.go
TimmyExogenous: The `metaInfo` isn't allowed to be empty, the check was added in the last commit. It seems like the check should be implemented in the front end to avoid invalid transactions. As for the contract, would It be enough to check it in the precompile implementation?

precompiles/assets/tx.go
TimmyExogenous: thanks

precompiles/assets/types.go
TimmyExogenous: Using the string type might be clearer when querying or setting values. However, using enum can reduce gas costs. This can be decided later when this field is actually used.

precompiles/assets/types.go
TimmyExogenous: Yeah, the asset registration will depend on the related client chain's registration.

precompiles/assets/types.go
TimmyExogenous: Yeah, I will add a check for this.

precompiles/assets/types.go
TimmyExogenous: The length of assetAddr will be checked to ensure it is not smaller than 20, so it is fine whether or not it is checked in the `ExocoreGateway`.

precompiles/assets/types.go
TimmyExogenous: It seems better not to allow the metaInfo to be empty if we provide this field to describe the client chain. This can ensure detailed and clearer information about the client chain for all users. Of course, this is a minor issue that won't have a significant impact no matter how it's handled.

precompiles/assets/tx.go
```diff
@@ -16,16 +18,20 @@ import (
...
+	MethodRegisterClientChain = "registerClientChain"
+
+	MethodRegisterToken = "registerToken"
 )
 
-// DepositAndWithdraw deposit and withdraw the client chain assets for the staker,
+// DepositOrWithdraw deposit and withdraw the client chain assets for the staker,
 // that will change the state in assets module.
-func (p Precompile) DepositAndWithdraw(
+func (p Precompile) DepositOrWithdraw(
```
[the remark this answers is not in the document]
TimmyExogenous: It seems not necessary.

precompiles/assets/types.go
adu-web3: ok, I'd better add this check in contract as well

precompiles/assets/types.go
adu-web3: this is different: even if the `assetAddr` is `address(0)`, the length of `assetAddr` would be 20

precompiles/assets/types.go
adu-web3: From design perspective, I think your question actually means: whether native module should totally trust in `ExocoreGateway` or native module should do defensive parameter validation regardless the implementation of `ExocoreGateway`? personally I prefer the native module not totally trusting the implementation of `ExocoreGateway` and check for params even if they might have been checked in `ExocoreGateway`

bwhour (approved):

precompiles/assets/types.go
TimmyExogenous: Yeah, the params will be checked in the native module no matter how it's implemented in `ExocoreGateway`.

precompiles/assets/types.go
TimmyExogenous: I see your point. It doesn't seem necessary to check these special addresses here. If we are checking address(0), what about other special addresses like address(1) or address(0xffff)? It might be better to include these checks when registering the assets, even if we intend to address them. I'm not sure if checking these special addresses during DApp development is a default specification. If it is, adding related address checks during registration might be more standard.

adu-web3: ok, we can skip this as not neccessary

PR #118, opened 2024-06-30
fix(oracle): set height to the next block of context when do recache

leonz789:
## Description

When a node restart, the oracle module will do refill the memory cache by replay some history blocks with executing the `PrepareNextRound`, `SealRound` process, the block height in context should be set to the corresponding history height.

## Changes
Notable changes:
- set block height of `context` to the corresponding history height when do recache
- refactor code when grow `roundID` with price unchanged
- add error log for case that `roundID` not match when append price for next round
----

Closes #XXX

## Summary by CodeRabbit

- **New Features**
  - Introduced fallback mechanisms for price aggregation failures, improving reliability in edge cases.
  - Enhanced round ID management in price aggregation to ensure continuity and accuracy.

- **Bug Fixes**
  - Improved error logging to assist in diagnosing issues more effectively.

- **Documentation**
  - Added comments in the code for better clarity and maintainability.

- **Refactor**
  - Renamed several functions for improved code readability and consistency.
  - Updated method calls in tests to align with new function names.

PR #122, opened 2024-07-04
fix(oralce):separate prepareRound from endBlock into beginBlock

leonz789:
## Description

Move `prepareRound` from `EndBlock` into `Beginblock`.
Currently we invoke `prepareRound` in the `EndBlock` of every block after all transaction executed. When there're `create-price` transactions in the first block since a node restart(like when a node restart from some height `x`, and some `create-price` messages are included in height `x` by other validators), the `aggregator` will be initialized/recached when the first message executed, this will consume extra gas which will conduct different gasUsed compared to other validators/nodes execution.
In this fix, we put the `prepareRound` into `BeginBlock`. In this way, avoid the initialization of `aggregator` be executed inside transaction.

----

Closes #XXX

## Summary by CodeRabbit

- **New Features**
  - Enhanced logging for better context during price proposal processing.
   
- **Refactor**
  - Renamed `PrepareRound` method to `PrepareRoundBeginBlock` across various functionalities to improve clarity in block calculation logic.
  
- **Bug Fixes**
  - Adjusted block height calculations to ensure accurate data handling in the aggregator context.
  
- **Chores**
  - Updated methods in `AppModule` to include additional logic for preparing and resetting the oracle round.

### DIFF
--- a/x/oracle/keeper/aggregator/context.go
+++ b/x/oracle/keeper/aggregator/context.go
@@ -218,7 +218,8 @@ func (agc *AggregatorContext) SealRound(ctx sdk.Context, force bool) (success []
 	return success, failed
 }
 
-func (agc *AggregatorContext) PrepareRound(ctx sdk.Context, block uint64) {
+// TODO: test to remove PrepareRound into BeginBlock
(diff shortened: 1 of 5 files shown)


leonz789:
merged into #118

PR #132, opened 2024-07-13
return default price in case of price-feeder fail

leonz789:
## Description

In case that before any valid price has been updated, all the rounds would be updated with 0 price, we return default price as 1 to operator.

----

Closes #XXX

## Summary by CodeRabbit

- **Bug Fixes**
  - Implemented a default price for tokens with zero value to ensure they are not stored as zero-priced assets.
  
- **New Features**
  - Updated default price decimal precision from 0 to 8, enhancing the accuracy of price representations and calculations throughout the application.

### DIFF
--- a/x/oracle/keeper/prices.go
+++ b/x/oracle/keeper/prices.go
@@ -65,6 +65,13 @@ func (k Keeper) GetSpecifiedAssetsPrice(ctx sdk.Context, assetID string) (types.
 		}, types.ErrGetPriceRoundNotFound.Wrapf("no valid price for assetID=%s", assetID)
 	}
 	v, _ := sdkmath.NewIntFromString(price.Price)
+	// for tokens really have 0 price, it should be removed from assets support, not here to provide zero price
+	if v.IsNil() || v.LTE(sdkmath.ZeroInt()) {
+		return types.Price{
+			Value:   sdkmath.NewInt(types.DefaultPriceValue),
+			Decimal: types.DefaultPriceDecimal,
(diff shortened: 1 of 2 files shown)


bwhour (approved):

mikebraver:
rebase on the latest branch to pass all github actions?

mikebraver (approved):

x/oracle/keeper/prices.go
```diff
@@ -99,6 +106,15 @@ func (k Keeper) GetMultipleAssetsPrices(ctx sdk.Context, assets map[string]inter
 			}
 		} else {
 			v, _ := sdkmath.NewIntFromString(price.Price)
+			// for tokens really have 0 price, it should be removed from assets support, not here to provide zero price
+			if v.IsNil() || v.LTE(sdkmath.ZeroInt()) {
+				info = info + assetID + " "
+				prices[assetID] = types.Price{
+					Value:   sdkmath.NewInt(types.DefaultPriceValue),
+					Decimal: types.DefaultPriceDecimal,
```
TimmyExogenous, line 114: I found that the `DefaultPriceDecimal` is 0, which means the default price would be 1 USD. This could incentivize a DDoS attack on the oracle to increase the value of assets priced below 1 USD. 
Should we set it to 6 to reduce this attack motivation. But we need to consider whether the initial voting power is enough when starting the local network through `local_node.sh` if we make this change.

TimmyExogenous (approved):

MaxMustermann2 (approved):

PR #133, opened 2024-07-13
check decimal from msg matches params

leonz789:
## Description

check decimal in msg match with decimal defined in oracle params
----

Closes #XXX

## Summary by CodeRabbit

- **New Features**
	- Implemented decimal validation for nested price structures to ensure consistency and accuracy.
	- Added a method to check decimal values for token feeders, enhancing data validation.

### DIFF
--- a/x/oracle/types/params.go
+++ b/x/oracle/types/params.go
@@ -191,3 +191,9 @@ func (p Params) CheckRules(feederID uint64, prices []*PriceSource) (bool, error)
 	// return true if no rule set, we will accept any source
 	return true, nil
 }
+
+func (p Params) CheckDecimal(feederID uint64, decimal int32) bool {
+	feeder := p.TokenFeeders[feederID]
+	token := p.Tokens[feeder.TokenID]
+	return token.Decimal == decimal
+}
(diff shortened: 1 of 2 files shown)


cloud8little:
Test passed, when decimals in the price is not equal to decimal configured in the oracle params, show "decimal not match" error.
```
10:40AM INF AfterEpochEnd minted successfully=20000000000000000000aexo module=x/exomint
10:40AM INF prepare for next oracle round of each tokenFeeder height=923 module=x/oracle
10:40AM INF price proposal failed error="decimal not match: invalid input create price" height=923 module=x/oracle
10:40AM INF add new round with previous price under fail aggregation module=x/oracle price=3410 roundID=90 tokenID=1
```

cloud8little (approved):

MaxMustermann2 (approved):

x/oracle/types/params.go
```diff
@@ -191,3 +191,9 @@ func (p Params) CheckRules(feederID uint64, prices []*PriceSource) (bool, error)
 	// return true if no rule set, we will accept any source
 	return true, nil
 }
+
+func (p Params) CheckDecimal(feederID uint64, decimal int32) bool {
+	feeder := p.TokenFeeders[feederID]
```
bwhour, line 197: can this feeder/token nil ?

bwhour (approved):

x/oracle/types/params.go
leonz789: This method is called after valid check, it should be fine

PR #160, opened 2024-08-11
update oracle create-price tx nonce check and gas consumption

leonz789:
## Notable Changes
- set nonce state to track nonce for per validator of every feederID, at most maxNonce of transactions are acceptable for each round.
  - reject tx out of price window
  - reject tx not signed by validator
  - reject tx has nonce not equal to currentNonce+1 or not in [1,2,3]
  - do all these check in `AnteHandler` to prevent these txs from being included by mempool
- set GasMeter to InfiniteGasMeterWithLimit
  - amount/gas value of create-price tx will not matter
  - tendermint mempool will use infiniteGasMeterWithLimt.Limit() to track that tx's gasWanted, we set it to 0 for create-price
- set priority to math.MaxInt64 for create-price, and currently we use FIFO tendermint mempool, so this ensure create-price tx will be included in mempool when received.

(*for currently application mempool we use noOpMempool*)

----

Closes #XXX

## Summary by CodeRabbit

- **New Features**
	- Enhanced nonce management in the AggregatorContext for improved security against replay attacks.
	- Introduced methods for nonce management within the Keeper interface, allowing for efficient manipulation and tracking.
	- Updated method to prepare rounds at the end of the block rather than the beginning, optimizing message handling.
	- Added Protocol Buffers messages for structured nonce management related to feeders and validators.
	- Implemented a gas consumption mechanism for transaction sizes to optimize resource usage.
	- Added a new method to retrieve validators, expanding the interface of the AggregatorContext.

- **Bug Fixes**
	- Corrected nonce validation logic to ensure stricter control over nonce usage.

- **Chores**
	- Streamlined initialization process for caches and aggregator contexts to improve efficiency during block processing.
	- Refactored gas deduction logic for oracle price transactions to enhance performance and clarity.


app/ante/cosmos/sigverify.go
```diff
@@ -569,16 +582,16 @@ func signatureDataToBz(data signing.SignatureData) ([][]byte, error) {
...
-	}
-	for _, msg := range msgs {
-		if _, ok := msg.(*oracletypes.MsgCreatePrice); ok {
-			continue
-		}
-		return false
-	}
-	return true
-}
+// func isOracleCreatePriceTx(tx sdk.Tx) bool {
```
cloud8little: you can delete this code snipet anyway.

x/oracle/keeper/aggregator/context.go
```diff
@@ -85,17 +85,13 @@ func (agc *AggregatorContext) sanityCheck(msg *types.MsgCreatePrice) error {
 		return errors.New("signer is not validator")
 	}
 
-	if msg.Nonce < 1 || msg.Nonce > common.MaxNonce {
-		return errors.New("nonce invalid")
-	}
-
 	// TODO: sanity check for price(no more than maxDetId count for each source, this should be take care in anteHandler)
```
cloud8little: is this TODO line still needed?

x/oracle/keeper/single.go
```diff
@@ -111,18 +107,21 @@ func recacheAggregatorContext(ctx sdk.Context, agc *aggregator.AggregatorContext
 		agc.SetParams(p)
 		setCommonParams(p)
 	} else {
+		prev := int64(0)
 		for ; from < to; from++ {
 			// fill params
-			prev := int64(0)
+			//			prev := int64(0)
```
cloud8little: delete this line plz.

leonz789: updated

x/oracle/keeper/aggregator/context.go
leonz789: updated

app/ante/cosmos/sigverify.go
leonz789: updated

x/oracle/keeper/single.go
leonz789: updated

cloud8little (approved):

app/ante/utils/oracle.go
```diff
@@ -0,0 +1,22 @@
...
+	if len(msgs) == 0 {
+		return false
+	}
+	for _, msg := range msgs {
+		if _, ok := msg.(*oracletypes.MsgCreatePrice); ok {
+			continue
+		}
+		return false
+	}
+	return true
```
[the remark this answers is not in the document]
bwhour: this loop seems better.

app/ante/cosmos/txsize_gas.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	params := cgts.ak.GetParams(ctx)
+
+	ctx.GasMeter().ConsumeGas(params.TxSizeCostPerByte*sdk.Gas(len(ctx.TxBytes())), "txSize")
+
+	// simulate gas cost for signatures in simulate mode
+	if simulate {
+		// in simulate mode, each element should be a nil signature
+		sigs, err := sigTx.GetSignaturesV2()
```
bwhour, line 56: we can pre-alloc memory for sig slice. 

> signers := sigTx.GetSigners()
sigs := make([]signing.SignatureV2, 0, len(signers))
sigs, err := sigTx.GetSignaturesV2()

app/ante/cosmos/txsize_gas.go
```diff
@@ -0,0 +1,118 @@
...
+			var pubkey cryptotypes.PubKey
+
+			acc := cgts.ak.GetAccount(ctx, signer)
+
+			// use placeholder simSecp256k1Pubkey if sig is nil
+			if acc == nil || acc.GetPubKey() == nil {
+				pubkey = simSecp256k1Pubkey
+			} else {
+				pubkey = acc.GetPubKey()
+			}
```
bwhour, line 77: we can use a seperate function to calacuate gas cost.

> func (decorator ConsumeTxSizeGasDecorator) calculateSignatureGas(ctx sdk.Context, pubkey cryptotypes.PubKey, params authtypes.Params) sdk.Gas

app/ante/cosmos/context.go
```diff
@@ -0,0 +1,91 @@
...
+// on gas provided and gas used.
+// CONTRACT: Must be first decorator in the chain
+// CONTRACT: Tx must implement GasTx interface
+type SetUpContextDecorator struct{}
+
+func NewSetUpContextDecorator() SetUpContextDecorator {
+	return SetUpContextDecorator{}
+}
+
+func (sud SetUpContextDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (newCtx sdk.Context, err error) {
```
bwhour, line 32: this  infinite gas transaction OracleCreatePriceTx, we need  add some log to track them in case that resources are misused, and transactions are rampant . how about the frequency of this transaction ?

x/oracle/keeper/nonce.go
```diff
@@ -0,0 +1,175 @@
...
+			n.NonceList = append(n.NonceList, v)
+		}
+		k.setNonce(store, n)
+	} else {
+		k.setNonce(store, nonce)
+	}
+}
+
+// AddZeroNonceItemForValidators init the nonce of a specific feederID for a set of validators
+func (k Keeper) AddZeroNonceItemWithFeederIDForValidators(ctx sdk.Context, feederID uint64, valdiators []string) {
```
bwhour: valdiators : typo 
can check validators firstly.

>     if len(validators) == 0 {
        return 
    }

app/ante/cosmos/context.go
```diff
@@ -0,0 +1,91 @@
...
+	// all transactions must implement GasTx
+	gasTx, ok := tx.(GasTx)
+	if !ok {
+		// Set a gas meter with limit 0 as to prevent an infinite gas meter attack
+		// during runTx.
```
TimmyExogenous, line 37: Should we consider the issue that a validator might send a price creation transaction that consumes too much gas, potentially affecting block propagation and CometBFT? The gas limit is usually a reasonable value set after considering network transmission time and consensus timeout settings.

app/ante/cosmos/txsize_gas.go
```diff
@@ -0,0 +1,118 @@
...
+
+func (cgts ConsumeTxSizeGasDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (sdk.Context, error) {
+	sigTx, ok := tx.(authsigning.SigVerifiableTx)
+	if !ok {
+		return ctx, sdkerrors.ErrTxDecode.Wrap("invalid tx type")
+	}
+
+	// Skip gas consumption if tx is an OracleCreatePriceTx
+	if anteutils.IsOracleCreatePriceTx(tx) {
+		return next(ctx, tx, simulate)
```
TimmyExogenous, line 46: Should we add a tx size limit for Oracle transactions to avoid the abovementioned issue?

app/ante/cosmos/sigverify.go
```diff
@@ -257,7 +259,7 @@ func OnlyLegacyAminoSigners(sigData signing.SignatureData) bool {
 }
 
 func (svd SigVerificationDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (newCtx sdk.Context, err error) {
-	if isOracleCreatePriceTx(tx) {
+	if utils.IsOracleCreatePriceTx(tx) {
 		// TODO: verify ed25519 signature for create-price message which is signed by consensusKey
```
TimmyExogenous: Can the price transaction now be submitted by any account? If yes, it seems like this check should be implemented before mainnet.

TimmyExogenous (approved):

app/ante/cosmos/sigverify.go
leonz789: > Can the price transaction now be submitted by any account? If yes, it seems like this check should be implemented before mainnet.

yes, in this PR it is ensured that tx singed by validator's consensuskey will be accept only

app/ante/cosmos/context.go
leonz789: This tx type is limited to create-price and that related gas consume is basically fixed. The bond is:
tx/block: num(validator) * maxNonce / interval(blocks).
eg. we have 21 validators and maxNonce set to 3, interval set to 100 blocks. then the result would be: 0.63tx/block in average. and the execution of tx:create-price is deterministic(which result anticipated gas consumption). Then we can rely on the number of tx to control the resource consumption.

app/ante/cosmos/context.go
leonz789: This tx type is limited to create-price and that related gas consume is basically fixed. The bond is:
tx/block: num(validator) * maxNonce / interval(blocks).
eg. we have 21 validators and maxNonce set to 3, interval set to 100 blocks. then the result would be: 0.63tx/block in average. and the execution of tx:create-price is deterministic(which result anticipated gas consumption). Then we can rely on the number of tx to control the resource consumption.

leonz789: We set create-price tx with the highest priority since it's basically a system service. So we limit this tx with nonce limitation not gas consumption(since it will finally fill the txpool with that priority)

x/oracle/keeper/nonce.go
leonz789: This is used to init for dogfood validators, and that validator set should not be empty

app/ante/cosmos/txsize_gas.go
leonz789: Explained as above.
And some more information:
1. we regard oracle service as a system service, so we want(and set) that tx with highest priority
2. in the above condition, no matter how much gas limitation we set on create-price, it can eventually fill up the tx-pool with no other limitation
3. we use nonce/round to limit this potential infinite resource consumption

x/oracle/keeper/single.go
```diff
@@ -138,6 +137,17 @@
...
 		}
+
+		for b, p = range recentParamsMap {
+			// use the latest params before the current block height
+			if b < to && b > prev {
+				agc.SetParams(p)
+				prev = b
+				setCommonParams(p)
+			}
+		}
```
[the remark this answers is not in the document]
leonz789: Only used for in-memory calculation, not state related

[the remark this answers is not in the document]
leonz789: Only used for in-memory calculation, not state related

app/ante/cosmos/sigverify.go
```diff
@@ -257,7 +259,7 @@ func OnlyLegacyAminoSigners(sigData signing.SignatureData) bool {
 }
 
 func (svd SigVerificationDecorator) AnteHandle(ctx sdk.Context, tx sdk.Tx, simulate bool, next sdk.AnteHandler) (newCtx sdk.Context, err error) {
-	if isOracleCreatePriceTx(tx) {
+	if utils.IsOracleCreatePriceTx(tx) {
```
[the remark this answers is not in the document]
leonz789: implemented, and removed this line

app/ante/utils/oracle.go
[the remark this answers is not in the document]
leonz789: updated

app/ante/cosmos/txsize_gas.go
leonz789: added limitation for txSize.

app/ante/cosmos/context.go
MaxMustermann2: But if the validator attempts to do a Sybil attack by sending many such transactions, how is it prevented?

x/oracle/keeper/nonce.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func (k Keeper) removeNonceWithValidator(store prefix.Store, validator string) {
+	store.Delete(types.NonceKey(validator))
+}
+
+func (k Keeper) removeNonceWithValidatorAndFeederID(store prefix.Store, validator string, feederID uint64) bool {
+	if nonce, found := k.getNonce(store, validator); found {
+		for i, n := range nonce.NonceList {
+			if n.FeederID == feederID {
```
MaxMustermann2, line 158: Consider optimizing this loop by using some sort of a special key / value combination. For example we could store the key to be `validator` + `feederID`.

MaxMustermann2 (approved):

PR #170, opened 2024-08-26
feat(oracle): support update for eth-native-restaking balance change

leonz789:
## Description

This PR add support for feed effective balance change of staked ETH on beaconChain onto exocoreChain.
When staked ETH on beaconChain deposit to exocoreChain, we treat it as a virtual LST and set it with a virtual price of its original amount of ETH on beaconChain.
eg.
validator_A has 32 staked ETH on beaconChain, and then deposit these 32 staked ETH to eoxcoreChain. He will get 32 NRETH(virtualLST) on exocoreChain, with price as 1ETH/NRETH. When validator_A got slashed on beaconChain which lead to his effective balance became to 16 ETH, then the price of his 32NRETH on exocoreChain will change to 16/32=0.5ETH/NRETH correspondingly.
*the symbol NRETH is used for clarify*

~TODO:~
~- [x] test~

----

Closes #XXX

## Summary by CodeRabbit

- **New Features**
	- Enhanced application functionality with the integration of the OracleKeeper for managing oracle data.
	- Introduced a new Protocol Buffers schema for token pricing and staking, improving data communication within the Exocore ecosystem.
	- Added methods for managing native tokens, including deposit, withdrawal, and delegation processes.
	- Implemented new functionality for retrieving asset IDs based on token IDs.
	- Introduced a comprehensive test suite for native token lifecycle management.

- **Bug Fixes**
	- Improved handling of native token operations to ensure accurate state management during deposits and withdrawals.

- **Documentation**
	- Updated comments and documentation to reflect new functionalities and enhancements related to native tokens and oracles.


x/operator/keeper/abci.go
```diff
@@ -62,6 +64,11 @@ func (k *Keeper) UpdateVotingPower(ctx sdk.Context, avsAddr string) error {
 			SelfUSDValue:   sdkmath.LegacyNewDec(0),
 			ActiveUSDValue: sdkmath.LegacyNewDec(0),
 		}
+		for _, assetID := range assetstypes.GetNativeTokenAssetIDs() {
+			if _, ok := prices[assetID]; ok {
+				prices[assetID], err = k.oracleKeeper.GetSpecifiedAssetsPrice(ctx, strings.Join([]string{assetID, operator}, "_"))
```
TimmyExogenous, line 69: should the returned error be addressed?

TimmyExogenous (commented):
It seems that this PR treats the native token as an independent token. It might be possible to treat the same native token from different stakers as a base token for processing. The oracle would only need to update the changes in the native token amount for each staker. When the amount increases, it would be equivalent to a deposit operation by the staker. When the amount decreases, it would be treated as a slash operation on the staker, similar to an immediate undelegation where the undelegated share gets slashed. This way, there would be no need to modify the existing processes for deposit, withdraw, delegation, undelegation, and updating voting power.

PR #157, opened 2024-08-04
feat: add flag to enable price-feeder on node start

leonz789:
## Description

<p> add flag `--oracle` to start price-feeder on exocored node start.</p>
----

Closes #XXX

## Summary by CodeRabbit

- **New Features**
	- Introduced new command-line flags for starting the Exocore node, allowing users to enable an oracle feeder and specify a mnemonic for the validator.

- **Improvements**
	- Enhanced control flow to allow asynchronous operation of the price feeder during node startup.
	- Added new dependencies to support price feeding functionality and updated various dependencies to improve stability and performance.

These changes will provide users with more flexibility in managing their nodes and ensure a better overall experience.

PR #174, opened 2024-08-29
auto add oracle configuration file in config file

cloud8little:
## Description

Update local_node.sh script to auto generate oracle related configuration file, and enable oracle when starting the node.

----

Dependency: 
#157  #160

Objective: 
Enable oracle for local_node test script, it will start feeding price at start block 20, update for every 10 blocks.


bwhour (approved):

MaxMustermann2 (approved):
Should the binary of the price feeder be installed? Where?

cloud8little:
> Should the binary of the price feeder be installed? Where?

No, the price feeder is a dependency here and exocore will call its start function inside.

leonz789 (approved):

PR #194, opened 2024-09-24
fix(audit): Fix issues from audit

TimmyExogenous:
## Description

This PR addresses the issues highlighted in the audit report. Only BTG-01 and URW-02 require code changes, while the others only need responses.

- [x] BTG-01: Chuang to require minimum self-delegation before opt-in (what about registration DoS?)
fix: Check whether the USD value owned by the operator is greater than or equal to this parameter when opting in.
answer: The registration DOS attack is mitigated by gas costs, similar to other transactions. Because the voting power will not be calculated if the registered operator has not opted into any AVS.

- [x] ENI-07: Chuang is checking
answer: The ratio won't be 1 if the operator is slashed because the slashing will decrease the staking amount of the asset pool while keeping the share unchanged.

- [x] URW-02: Chuang is checking
fix: The mechanism of updating the assets state in Exocore chain by event log has been deprecated, I will remove the related code.

- [x] EUK-01: Part of slashing implementation: Chuang?
answer: It's unused now, currently we only support the slashing from dogfood. I will disable the precompile.

- [x] URW-03: Chuang
answer: The slashing function in the slash module is deprecated, now we use the `slash.go` in the operator module to address the slashing.


TimmyExogenous:
All issues have been answered through Skyharbor.

trestinlsd (approved):
look good

mikebraver (approved):

precompiles/avs/avs_test.go
```diff
@@ -1,6 +1,9 @@
 package avs_test
 
 import (
+	"cosmossdk.io/math"
+	types2 "github.com/ExocoreNetwork/exocore/x/assets/types"
```
bwhour: rename this types2 to assetstypes

x/avs/keeper/avs_test.go
```diff
@@ -1,6 +1,9 @@
 package keeper_test
 
 import (
+	"cosmossdk.io/math"
+	types2 "github.com/ExocoreNetwork/exocore/x/assets/types"
```
bwhour: rename this types2 to assetstypes

x/operator/keeper/opt.go
```diff
@@ -30,6 +29,20 @@ func (k *Keeper) OptIn(ctx sdk.Context, operatorAddress sdk.AccAddress, avsAddr
 	if k.IsOptedIn(ctx, operatorAddress.String(), avsAddr) {
 		return types.ErrAlreadyOptedIn
 	}
+	// Check if the USD value of the operator is greater than or equal to the self-delegation
+	// configured by the AVS. This is used to prevent a DDOS attack from zero-USD value opting in.
+	operatorUSDValues, err := k.GetOrCalculateOperatorUSDValues(ctx, operatorAddress, avsAddr)
+	if err != nil {
+		return errorsmod.Wrapf(err, "OptIn: error when calculating operator USD value, operator:%s avsAddr:%s", operatorAddress.String(), avsAddr)
+	}
+	minSelfDelgation, err := k.avsKeeper.GetAVSMinimumSelfDelegation(ctx, avsAddr)
```
bwhour: typo: minSelfDelegation

TimmyExogenous: thanks

precompiles/avs/avs_test.go
TimmyExogenous: thanks

x/avs/keeper/avs_test.go
TimmyExogenous: thanks

bwhour (approved):
lgtm.

PR #201, opened 2024-09-27
deal with corner case that balance be negative

leonz789:
## Description
When effective balances are updated from transactions by validators, there's one case that this delta might be more than previous Balance:
1. staker's validatorlist: {v1, v2, v3, v5}
2. in one same block: withdraw v2, v3, v5, balance of v2, v3, v5 all be slashed by -16
3. => amount: 32*4->32(by withdraw), the validatorList of feeder will be updated on next block, so it will report the balance change of v5: -16 as in the staker's balance change, result to: 32*4->32-> 32-16*3 = -16
we will just ingore this misbehavior introduced by synchronize-issue, and this will be correct in next block/round
## Notable Changes
- when balance is changed to be negative, skip this update
- update event field when deposit/withdraw from beacon chain happened. Use index instead of length of validator list to make sure that price-feeder is consistent with exocored on staker's validator list

----

Closes #XXX

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -163,7 +163,7 @@ require (
 	github.com/holiman/uint256 v1.2.4 // indirect
 	github.com/huandu/skiplist v1.2.0 // indirect
 	github.com/huin/goupnp v1.3.0 // indirect
-	github.com/imroc/biu v0.0.0-20170329141542-0376ce6761c0 // indirect
+	github.com/imroc/biu v0.0.0-20170329141542-0376ce6761c0
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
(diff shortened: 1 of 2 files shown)


TimmyExogenous (approved):
LGTM.
I have a question, do you think whether it's necessary to add a `deltaMaxLimit` to limit an extremely large delta value, taking into account the deviation caused by the timeliness of state updates. However, I'm unsure how large this value should be set, as it could depend on the short-term increase in the number of client chain validators from stakers, the impact of slashing and rewards on the client chain side, and of course, the timeliness of oracle state updates.

MaxMustermann2 (approved):

bwhour (approved):
looks good.

cloud8little (approved):

PR #259, opened 2024-12-05
fix: genesis data validation

MaxMustermann2:
1. In x/assets, non-Ethereum assets are now supported.
2. The `total_staking_amount` for the `exo` token is not recorded, and hence, the validation fails. Instead, this should be statefully checked.
3. Zero slashing is exported, so importing should be allowed too. The other alternative is to disallow 0 slashing export.

## Release Notes

- **New Features**
  - Enhanced asset management with improved initialization of key components, including `AssetsKeeper`, `DelegationKeeper`, and `StakingKeeper`.
  - Introduced a new `AccountKeeper` interface for better account management.
  - Added new CLI command for querying associated stakers by operator.
  - Implemented a method to retrieve associated stakers for a specified operator.
  - Expanded delegation module functionality with new query messages and RPC method.

- **Bug Fixes**
  - Improved validation logic for asset delegation during initialization to ensure accurate financial data.
  - Adjusted slashing state validations to allow zero values while maintaining checks for nil or negative amounts.
  - Added duplicate asset ID checks in asset validation.

- **Documentation**
  - Updated comments and documentation to clarify changes and functionality across various components.


cloud8little:
Testing.

MaxMustermann2:
Query test using `local_node.sh`

Command
```bash
exocored query delegation QueryAssociatedStakersByOperator \
    exo18cggcpvwspnd5c6ny8wrqxpffj5zmhklprtnph --output json | jq
```

Response
```json
{
  "stakers": [
    "0x3e108c058e8066da635321dc3018294ca82ddedf_0x65"
  ]
}
```

cloud8little (approved):

bwhour (approved):
looks good.

x/delegation/keeper/delegation_state.go
```diff
@@ -373,6 +373,27 @@ func (k *Keeper) GetAssociatedOperator(ctx sdk.Context, stakerID string) (string
...
 
+func (k *Keeper) GetAssociatedStakers(ctx sdk.Context, operator string) ([]string, error) {
+	if _, err := sdk.AccAddressFromBech32(operator); err != nil {
+		return nil, delegationtype.ErrOperatorAddrIsNotAccAddr
+	}
+	store := prefix.NewStore(ctx.KVStore(k.storeKey), delegationtype.KeyPrefixAssociatedOperatorByStaker)
+	iterator := sdk.KVStorePrefixIterator(store, []byte{})
+	defer iterator.Close()
+
+	// assuming that we support 5 client chains, this is a reasonable capacity.
```
TimmyExogenous, line 384: Q: 
does this mean we only allow one staker from a client chain to associate with an operator?  
I didn’t impose this restriction when implementing the association function.  
Do you think we need to enforce this restriction in the association function?

TimmyExogenous (approved):
LGTM

x/delegation/keeper/delegation_state.go
MaxMustermann2: Fair point. We don't need this restriction. But would you still advise a different starting capacity?

TimmyExogenous: No, I’m just not sure where this RPC will be used or whether it will impact the caller if multiple stakers from one client chain associate with a single operator.

PR #284, opened 2025-01-12
feat(delegation): auto associate when possible

MaxMustermann2:
Whenever the asset bearing `ExocoreAssetID` is delegated by a staker to an operator, we should check if the staker address is the same as the operator address. If a match is found, they should be automatically associated.

## Release Notes

- **New Features**
  - Added automatic operator-staker association for self-delegation scenarios
  - Enhanced delegation process with improved association mechanism

- **Testing**
  - Added new test case to validate automatic operator association during self-delegation
  - Updated test suite with refined asset and constant references

- **Improvements**
  - Updated module dependencies to utilize Exocore network types
  - Added clarifying comments for asset information handling related to slashing and undelegation

### DIFF
--- a/x/delegation/keeper/delegation.go
+++ b/x/delegation/keeper/delegation.go
@@ -1,6 +1,7 @@
 package keeper
 
 import (
+	"bytes"
 	"fmt"
 
 	errorsmod "cosmossdk.io/errors"
@@ -61,6 +62,13 @@ func (k *Keeper) delegateTo(
 		// transfer the delegation amount from the staker account to the delegated pool
 		if err := k.bankKeeper.DelegateCoinsFromAccountToModule(ctx, params.StakerAddress, delegationtype.DelegatedPoolName, coins); err != nil {
 			return err
+		}
+		// auto associate it, if there is a match. note that both are byte versions of bech32 AccAddress.
+		if bytes.Equal(params.StakerAddress, params.OperatorAddress[:]) {
+			err := k.SetAssociatedOperator(ctx, stakerID, params.OperatorAddress.String())
+			if err != nil {
+				return err
+			}
 		}
 	}
 	// calculate the share from the delegation amount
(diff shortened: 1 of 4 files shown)


x/delegation/keeper/delegation.go
```diff
@@ -62,6 +63,13 @@ func (k *Keeper) delegateTo(
 		if err := k.bankKeeper.DelegateCoinsFromAccountToModule(ctx, params.StakerAddress, delegationtype.DelegatedPoolName, coins); err != nil {
 			return err
 		}
+		// auto associate it, if there is a match. note that both are byte versions of bech32 AccAddress.
+		if bytes.Equal(params.StakerAddress, params.OperatorAddress[:]) {
+			err := k.SetAssociatedOperator(ctx, stakerID, params.OperatorAddress.String())
```
TimmyExogenous, line 75: Can we check if it has already been associated to avoid repetitive association?

x/delegation/keeper/delegation.go
```diff
@@ -62,6 +63,13 @@ func (k *Keeper) delegateTo(
 		if err := k.bankKeeper.DelegateCoinsFromAccountToModule(ctx, params.StakerAddress, delegationtype.DelegatedPoolName, coins); err != nil {
 			return err
 		}
+		// auto associate it, if there is a match. note that both are byte versions of bech32 AccAddress.
+		if bytes.Equal(params.StakerAddress, params.OperatorAddress[:]) {
```
adu-web3, line 73: do we still need the associate/dissociate function in delegation precompile in this case? since the evm address that initiates the client chain delegation request could share the same bytes representation with the operator address, and we can require that operator should initiate delegations request from client chain with an evm address sharing the same bytes representation with its exocore bech32 operator address(meaning the same private key)

MaxMustermann2: I thought the intention for the precompile's association function is to support it on non EVM chains

adu-web3 (approved):

x/delegation/keeper/delegation.go
MaxMustermann2: Updated the comment as discussed; please check it out.

TimmyExogenous (approved):

PR #307, opened 2025-02-17
feat(gov): disable the governance through set null router

TimmyExogenous:
## Description

Max has already added the authority check for all modules when updating parameters. So, this PR only disables governance by setting the routers to null, including both `MsgServiceRouter` and `legacyRouter`.

During the early phase of mainnet, the parameters can be updated through a node upgrade without enabling the governance module. An example of updating the gateways is provided in this documentation:
[updating gateways by node upgrade](https://www.notion.so/exocore/upgrade-and-cosmovisor-182612d2b5df80b7bd19d89d7dfbe5d1?pvs=4#182612d2b5df806a9e20f1b5c5b53232)

Additionally, I am unsure whether we should provide a precompile interface to update the list of assets supported by Dogfood. This interface should be controlled by the owner of the gateway contract, similar to the `registerToken`. It would allow us to add supported assets for the Dogfood AVS without upgrading the node. I can add it in this PR if we think it's necessary.

## Summary by CodeRabbit

- **Refactor**
  - Temporarily disabled governance functionality during the early mainnet phase.
  - Updated message routing within the governance module to align with the transitional state.

These changes are designed to facilitate the platform's adaptation to the evolving mainnet environment.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -88,9 +88,7 @@ import (
 	ibctransfer "github.com/cosmos/ibc-go/v7/modules/apps/transfer"
 	ibctransfertypes "github.com/cosmos/ibc-go/v7/modules/apps/transfer/types"
 	ibc "github.com/cosmos/ibc-go/v7/modules/core"
-	ibcclient "github.com/cosmos/ibc-go/v7/modules/core/02-client"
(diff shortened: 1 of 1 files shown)


MaxMustermann2 (approved):

bwhour (approved):
looks good.

leonz789 (approved):

PR #325, opened 2025-03-17
refact(bls):implement gas calculation logic

trestinlsd:
## Description

----

## Summary by CodeRabbit

- **New Features**
	- Introduced dynamic gas cost calculations with updated resource cost parameters for cryptographic operations.
- **Refactor**
	- Standardized naming conventions and updated interface declarations for improved clarity and consistency.
	- Added a new file defining gas constants for BLS12-381 elliptic curve operations.
- **Tests**
	- Adjusted test cases to validate the updated naming and gas computation logic.
- **Chore**
	- Removed legacy configuration parameters to simplify initialization and streamline operation setup.

### DIFF
--- a/precompiles/bls/bls.go
+++ b/precompiles/bls/bls.go
@@ -9,8 +9,6 @@ import (
 	"golang.org/x/crypto/sha3"
 
 	"github.com/cometbft/cometbft/libs/log"
-	imuacmn "github.com/imua-xyz/imuachain/precompiles/common"
-
 	sdk "github.com/cosmos/cosmos-sdk/types"
 
 	"github.com/ethereum/go-ethereum/accounts/abi"
@@ -29,12 +27,11 @@ var f embed.FS
 // Precompile defines the precompiled contract for deposit.
 type Precompile struct {
 	abi.ABI
-	baseGas uint64
 }
 
 // NewPrecompile creates a new BLS Precompile instance as a
 // PrecompiledContract interface.
-func NewPrecompile(baseGas uint64) (*Precompile, error) {
+func NewPrecompile() (*Precompile, error) {
 	abiBz, err := f.ReadFile("abi.json")
 	if err != nil {
 		return nil, fmt.Errorf("error loading the deposit ABI %s", err)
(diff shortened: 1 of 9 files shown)


precompiles/bls/bls.go
```diff
@@ -56,9 +61,84 @@ func (p Precompile) Address() common.Address {
...
+
 // RequiredGas calculates the precompiled contract's base gas rate.
-func (p Precompile) RequiredGas(_ []byte) uint64 {
-	return p.baseGas
+func (p Precompile) RequiredGas(input []byte) uint64 {
+	if len(input) < 4 {
+		return 0
+	}
+	selector := input[:4]
+	switch {
```
MaxMustermann2: Why not use `p.MethodById` directly ?

precompiles/common/protocol_params.go
```diff
@@ -0,0 +1,13 @@
+package common
```
MaxMustermann2: Move this file to the `bls`  subfolder.

precompiles/bls/bls.go
```diff
@@ -4,8 +4,13 @@ import (
 	"bytes"
```
MaxMustermann2, line 4: `BaseGas` within the `type Precompile struct` is not used.

MaxMustermann2 (changes requested):

precompiles/bls/bls.go
trestinlsd: It cannot be used directly. p.MethodById is usually used as follows: 
methodID := input[:4]
method, err := p.MethodById(methodID)
ID :=method.ID

precompiles/bls/bls.go
trestinlsd: fixed

precompiles/bls/bls.go
MaxMustermann2: Yes; so you will not need to generate selectors. It will do that on your behalf.

precompiles/bls/bls.go
```diff
@@ -56,9 +53,85 @@ func (p Precompile) Address() common.Address {
...
+
+	case bytes.Equal(selector, aggregateSignaturesSelector):
+		return p.calculateAggregationGas(input, Bls12381G2AddGas)
+
+	case bytes.Equal(selector, addTwoPubKeysSelector):
+		return Bls12381G1AddGas
+
+	default:
+		return 0
+	}
```
MaxMustermann2, line 82:
```go
if len(input) < 4 {
    return 0
}
method, err := p.MethodById(input)
if err != nil {
   return 0
}
switch method.Name {
case MethodFastAggregateVerify:
    return  Bls12381PairingBaseGas + Bls12381PairingPerPairGas
// so on
}
```

cary007 (approved):

MaxMustermann2 (approved):

PR #335, opened 2025-04-28
Feat/2 phases delayversion

leonz789:
## Description
The price feeder submits NST balance changes and uses a ‘version’ field to ensure synchronization. However, if the ‘version’ changes during the two-phase process of submitting a price, it causes a ‘version mismatch’, which breaks the update process.
So we introduced a new version type called ‘feedVersion’. We also added the accumulated deposit amount along with a version field to prevent the update process from breaking due to deposits during the two-phase submission.

e.g
```
{
  Version {9, 1200}
  FeedVersion {7, 1000}
}
```
At this point in the state:
The price feeder will submit balance based on FeedVersion 7, which corresponds to the staker list at version 7. When the Oracle module receives the balance change, it will calculate the adjusted total as: `total - (1200 - 1000)`
This adjustment removes the impact of newly added deposit amounts (validators) since the price feeder did not fetch balances for validators added after version 7.

After a successful submission and update, the versions will advance to:
```
{
  Version: {10, 1200}
  FeedVersion: {10, 1200}
}
```
If a new deposit happens, it will only increment Version and its associated accumulated deposit amount, but not FeedVersion.
```
{
  Version: {11, 1300}
  FeedVersion: {10, 1200}
}
```

----

Closes #XXX

## Summary by CodeRabbit

- **New Features**
  - Implemented a two-phase oracle data submission protocol with Merkle proof verification.
  - Added support for Solana-like chain (SOLChain) and native tokens (NSTETH, NSTSOL).
  - Extended protobuf schemas for NST staker info, balances, and Merkle trees.
  - Developed a custom mempool for raw data caching and selection.
  - Introduced new event attributes for NST updates and malicious reports.

- **Enhancements**
  - Improved NST staker management with versioning, indexing, and balance tracking.
  - Enhanced event emission and query responses for NST data.
  - Refined validation for oracle message types and two-phase flows.
  - Added detailed logging and error handling in keeper and message processing.
  - Upgraded feeder management with slashing, jailing, and nonce logic for two-phase.
  - Expanded test coverage for two-phase aggregation, Merkle proofs, and malicious scenarios.
  - Clarified interfaces, methods, and documentation.
  - Prevented mixing of create-price messages with other tx messages.
  - Improved message encoding and event attribute semantics.
  - Extended keeper cache and recovery for two-phase state.

- **Bug Fixes**
  - Corrected parameter types and API protobuf definitions.
  - Fixed staker index and balance calculation logic.
  - Ensured create-price messages are isolated from other message types.

- **Chores**
  - Updated dependencies for new chains/tokens; removed deprecated code.
  - Reorganized code for clarity, modularity, and maintainability.
  - Added static analysis suppressions for safe casts.
  - Commented out disabled external feeder initialization code.


leonz789:
superseded by #353

PR #342, opened 2025-05-19
BREAKING: feat(undelegation): add instant unbonding

mikebraver:
## Description
This pr add instant unbonding for staker undelegation:
If unbonding period is more than 1 epoch away:
-  Apply 25% penalty
- Process unbonding immediately
- Set completion to current block
If unbonding period is 1 epoch or less:
- No penalty
- Use normal unbonding period
- Process after unbonding period completes

----

## Summary by CodeRabbit

- **New Features**
  - Added support for instant undelegation with a configurable penalty.
  - Introduced new API endpoints and CLI commands to query and update instant undelegation parameters.
  - Added new events and messages to support instant unbonding workflows.

- **Improvements**
  - Enhanced undelegation logic to distinguish between instant and regular unbonding processes.
  - Enabled governance authority to update instant undelegation penalty parameters via messages.

- **Bug Fixes**
  - Updated test utilities for improved error handling consistency.

- **Documentation**
  - Expanded API and parameter documentation to cover instant undelegation functionality.


x/delegation/keeper/delegation.go
```diff
@@ -158,6 +160,32 @@ func (k *Keeper) UndelegateFrom(ctx sdk.Context, params *delegationtype.Delegati
...
+
+	// Get current epoch info
+	currentEpochInfo, found := k.epochsKeeper.GetEpochInfo(ctx, completedEpochID)
+	if !found {
+		return fmt.Errorf("epoch info not found for identifier: %s", completedEpochID)
+	}
+
+	// Determine if this should be an instant unbonding
+	// If the unbonding period is more than 1 epoch away, treat it as instant unbonding
+	isInstant := completedEpochNumber-currentEpochInfo.CurrentEpoch > 1
```
TimmyExogenous: This implementation means that all undelegations with an unbonding duration longer than one epoch will be forcibly converted to instant undelegations. If that's the case, the unbonding duration becomes meaningless.

It might be better to provide a separate function for instant undelegation, allowing stakers to choose between instant undelegation with some penalty and normal undelegation with a longer unbonding period but no penalty.

TimmyExogenous (commented):
Instant undelegation might affect voting power updates and reward allocation. These effects may need to be handled carefully if we require the undelegation to take effect immediately at the block where it is submitted.

However, if we allow the instant undelegation to take effect at the end of the current epoch, the handling becomes simpler, since both voting power updates and reward distribution are processed on a per-epoch basis.

x/delegation/keeper/delegation.go
mikebraver: Updated

x/delegation/keeper/delegation.go
```diff
@@ -132,10 +133,10 @@ func (k *Keeper) delegateTo(
 	return nil
 }
 
-// UndelegateFrom: The undelegation needs to consider whether the operator's opted-in assets can exit from the AVS.
-// Because only after the operator has served the AVS can the staking asset be undelegated.
-// So we use two steps to handle the undelegation. Fist,record the undelegation request and the corresponding exit time which needs to be obtained from the operator opt-in module. Then,we handle the record when the exit time has expired.
-func (k *Keeper) UndelegateFrom(ctx sdk.Context, params *delegationtype.DelegationOrUndelegationParams) error {
+// UndelegateFrom handles both normal and instant undelegation
+// If isInstant is true, it will apply a penalty if the current epoch is less than the completion epoch
+// If isInstant is false, it will apply a penalty if the unbonding expiration is more than 1 epoch away
```
TimmyExogenous: Why is there a penalty if `isInstant` is false?

x/delegation/keeper/delegation.go
```diff
@@ -158,23 +159,48 @@ func (k *Keeper) UndelegateFrom(ctx sdk.Context, params *delegationtype.Delegati
...
+		StakerId:                 stakerID,
+		AssetId:                  assetID,
+		OperatorAddr:             params.OperatorAddress.String(),
+		TxHash:                   params.TxHash.String(),
+		UndelegationId:           undelegationID,
+		BlockNumber:              uint64(ctx.BlockHeight()),
+		Amount:                   removeToken,
+		ActualCompletedAmount:    actualCompletedAmount,
+		CompletedEpochIdentifier: completedEpochID,
+		CompletedEpochNumber:     completedEpochNumber,
```
TimmyExogenous: It looks like the undelegation is always completed when the unbonding duration expires, regardless of whether `isInstant` is true or false. This implementation only applies a penalty when `isInstant` is true and the unbonding duration is less than one epoch.

precompiles/delegation/IDelegation.sol
```diff
@@ -40,12 +40,14 @@ interface IDelegation {
...
 /// @param opAmount The Undelegation amount
+/// @param isInstant Whether to perform instant undelegation with penalty
     function undelegate(
         uint32 clientChainID,
         bytes calldata assetsAddress,
         bytes calldata stakerAddress,
         bytes calldata operatorAddr,
-        uint256 opAmount
+        uint256 opAmount,
+        bool isInstant
```
TimmyExogenous: This update modifies the precompile interface, which may introduce compatibility issues for the gateway contract. @adu-web3

x/delegation/keeper/delegation.go
```diff
@@ -158,23 +159,48 @@ func (k *Keeper) UndelegateFrom(ctx sdk.Context, params *delegationtype.Delegati
...
+	if !found {
+		return fmt.Errorf("epoch info not found for epoch ID: %s", completedEpochID)
+	}
+
+	// Calculate actual completed amount based on undelegation type
+	var actualCompletedAmount sdkmath.Int
+	if isInstant {
+		// For instant undelegation, apply penalty if current epoch is less than completion epoch
+		if epochInfo.CurrentEpoch < completedEpochNumber {
+			penaltyAmount := removeToken.ToLegacyDec().Mul(sdkmath.LegacyNewDecWithPrec(25, 2)).TruncateInt()
```
TimmyExogenous: Should we configure the penalty ratio as a parameter in the delegation module so that it can be set via genesis or governance in the future?
If not, please define a constant value for now.

rlan35:
> Instant undelegation might affect voting power updates and reward allocation. These effects may need to be handled carefully if we require the undelegation to take effect immediately at the block where it is submitted.
> 
> However, if we allow the instant undelegation to take effect at the end of the current epoch, the handling becomes simpler, since both voting power updates and reward distribution are processed on a per-epoch basis.

Let's do it at the end of current epoch. Instant doesn't mean it happens right away, it means at current epoch.

precompiles/delegation/IDelegation.sol
adu-web3: we need to update gateway contract accordingly in this case

precompiles/delegation/tx.go
```diff
@@ -85,7 +85,17 @@ func (p Precompile) Undelegate(
 	}
 	undelegationParams.TxHash = txHash
 
-	err = p.delegationKeeper.UndelegateFrom(ctx, undelegationParams)
+	// Get instantUnbond parameter
+	instantUnbond, ok := args[5].(bool)
```
leonz789: Integrate this into `GetDelegationParamsFromInputs` ?

proto/imuachain/delegation/v1/query.proto
```diff
@@ -49,6 +49,9 @@ message DelegationAmounts {
     (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
     (gogoproto.nullable) = false
   ];
+
+  // instant_unbond indicates whether to perform instant unbonding with penalty
```
leonz789: 'whether this is a instant unbonding with penalty' ?

proto/imuachain/delegation/v1/tx.proto
```diff
@@ -16,6 +16,10 @@ message ValueField {
     (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Int",
     (gogoproto.nullable) = false
   ];
+  // instant_unbond indicates whether to perform instant unbonding with a 25% penalty.
+  // If true, tokens will be returned immediately with a 25% penalty.
+  // If false, tokens will be returned after the waiting period with no penalty.
+  bool instant_unbond = 2;
```
leonz789: Maybe add a new filed for 'MsgUndelegation' ?
'ValueField' is meant to wrap 'int' value, there's also a 'DecValueFiled'

x/delegation/client/cli/tx_delegation_native.go
```diff
@@ -55,12 +55,36 @@ func CmdUndelegate() *cobra.Command {
...
+
+			// Create the message
+			msg := types.NewMsgUndelegation(
+				assetID,
+				clientCtx.GetFromAddress().String(),
+				[]types.KeyValue{{
+					Key: operatorAddrStr,
+					Value: &types.ValueField{
+						Amount:        amount,
+						InstantUnbond: instantUnbond,
```
leonz789: set this as a filed of `msgUndelegation` instead of `Value` ?

x/delegation/keeper/msg_server.go
```diff
@@ -70,8 +70,15 @@ func (k *Keeper) UndelegateAssetFromOperator(
 	)
 	cachedCtx, writeFunc := ctx.CacheContext()
 	for _, inputParams := range inputParamsList {
-		if err := k.UndelegateFrom(cachedCtx, inputParams); err != nil {
-			return nil, err
+		// Get instant unbonding flag from the message
+		if msg.BaseInfo.PerOperatorAmounts[0].Value.InstantUnbond {
```
leonz789: please put the `InstantUnbond` flag in message level

x/delegation/keeper/msg_server.go
```diff
@@ -70,8 +70,15 @@ func (k *Keeper) UndelegateAssetFromOperator(
...
+			if err := k.InstantUndelegateFrom(cachedCtx, inputParams); err != nil {
```
leonz789, line 84: `inputParamsList := newDelegationParms(...)`
the method `newDelegationParams` set type to be fixed: `delegateTo`, but here should be `undelegateFrom`

proto/imuachain/delegation/v1/query.proto
leonz789: Also, DelegationAmounts is a summarized total, not tied to individual UndelegationRecords.
Therefore, it may not be reasonable for this message to include an instant flag, since the wait_undelegation_amount could be a combined total of both instant and non-instant undelegations.

precompiles/delegation/tx.go
mikebraver: `GetDelegationParamsFromInputs` is more about check against the genesis file, and for external command or precompiled way, we still need to check the passed args

x/delegation/keeper/delegation.go
```diff
@@ -132,6 +132,105 @@ func (k *Keeper) delegateTo(
...
+		OperatorAddr:          params.OperatorAddress.String(),
+		TxHash:                params.TxHash.String(),
+		UndelegationId:        undelegationID,
+		BlockNumber:           uint64(ctx.BlockHeight()),
+		Amount:                removeToken,
+		ActualCompletedAmount: removeToken,
+	}
+
+	if completedEpochNumber-epochInfo.CurrentEpoch > 1 {
+		r.CompletedEpochNumber = epochInfo.CurrentEpoch
```
leonz789: `r.CompletedEpochNumber = epochInfo.CurrentEpoch + 1`
We need to increment the epoch number here. Otherwise:

If block 100 is the last block of the current epoch, and we are currently at block 100:
	1.	During the transaction, the dogfood module’s hooks will call
wrapper.keeper.AppendUndelegationToMature(ctx, unbondingCompletionEpoch, recordKey)
using the current epoch (epochInfo.CurrentEpoch) as the key.
	2.	However, in the next BeginBlock, the epochs module invokes the dogfood hooks to populate pendingUndelegations,
which retrieves records by epoch key generated from the current epoch.
This causes the record set during the tx (in block 100) to be missed,
since by then, epochInfo.CurrentEpoch has already advanced to the next epoch.

precompiles/delegation/tx.go
leonz789: I mean this one 
`func (p Precompile) GetDelegationParamsFromInputs(ctx sdk.Context, args []interface{}) (*delegationtypes.DelegationOrUndelegationParams, error)`, with expanded `delegationtypes.DelegationOrUndelegationParams`

leonz789 (approved):

proto/imuachain/delegation/v1/genesis.proto
```diff
@@ -23,6 +24,8 @@ message GenesisState {
   repeated UndelegationAndHoldCount undelegations = 4 [(gogoproto.nullable) = false];
   // last_undelegation_id is a globally incremented ID assigned to undelegations.
   uint64 last_undelegation_id = 5;
+  // params refers to the parameters of the module.
+  InstantUndelegationPenaltyParams instant_undelegation_penalty_params = 6 [(gogoproto.nullable) = false];
```
TimmyExogenous: To keep the code style consistent with other modules, would it be better to rename it to `Params`, representing the parameters of the delegation module? In the future, we could also add more fields to accommodate potential requirements.

x/delegation/keeper/delegation.go
```diff
@@ -132,6 +132,105 @@ func (k *Keeper) delegateTo(
...
+	if !found {
+		return errorsmod.Wrapf(delegationtype.ErrEpochIdentifierNotExist, "identifier:%s", completedEpochID)
+	}
+
+	// Apply instant penalty only when the completion epoch is more than one epoch away
+	if completedEpochNumber-epochInfo.CurrentEpoch > 1 {
+		// Get the instant undelegation penalty
+		penalty := k.GetInstantUndelegationPenalty(ctx)
+		penaltyAmount := params.OpAmount.Mul(sdk.NewInt(int64(penalty))).Quo(sdk.NewInt(100))
+		params.OpAmount = params.OpAmount.Add(penaltyAmount)
```
TimmyExogenous: This means the penalty won’t be deducted from the input opAmount, which could cause an issue if the resulting opAmount (including the penalty) exceeds the total delegated amount.
It might be fine to change this line to:
```go
ActualCompletedAmount = params.OpAmount.Sub(penaltyAmount)
```

proto/imuachain/delegation/v1/genesis.proto
MaxMustermann2: Additionally, `params` should be the first element on the `genesis` state.

proto/imuachain/delegation/v1/params.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+
+package imuachain.delegation.v1;
+
+option go_package = "github.com/imua-xyz/imuachain/x/delegation/types";
+
+// Params defines the parameters for the module.
+message InstantUndelegationPenaltyParams {
```
MaxMustermann2:
```suggestion
message Params {
```

proto/imuachain/delegation/v1/query.proto
```diff
@@ -170,6 +171,15 @@ message QueryDelegatedStakersByOperatorResponse {
   repeated string stakers = 2;
 }
 
+// QueryParamsRequest is request type for the Query/Params RPC method.
+message QueryInstantUndelegationPenaltyParamsRequest {}
```
MaxMustermann2:
```suggestion
message QueryParamsRequest {}
```

proto/imuachain/delegation/v1/query.proto
```diff
@@ -170,6 +171,15 @@ message QueryDelegatedStakersByOperatorResponse {
...
+
+// QueryParamsResponse is response type for the Query/Params RPC method.
+message QueryInstantUndelegationPenaltyParamsResponse {
+  // params holds all the parameters of this module.
+  InstantUndelegationPenaltyParams instant_undelegation_penalty_params = 1 [(gogoproto.nullable) = false];
+}
```
MaxMustermann2:
```suggestion
// QueryParamsResponse is response type for the Query/Params RPC method.
message QueryParamsResponse {
  // params holds all the parameters of this module.
  Params params = 1 [(gogoproto.nullable) = false];
}
```

x/delegation/keeper/delegation.go
```diff
@@ -132,6 +132,105 @@ func (k *Keeper) delegateTo(
...
+			sdk.NewAttribute(delegationtype.AttributeKeyAssetID, r.AssetId),
+			sdk.NewAttribute(delegationtype.AttributeKeyOperator, r.OperatorAddr),
+			sdk.NewAttribute(delegationtype.AttributeKeyRecordID, hexutil.Encode(recordKey)),
+			sdk.NewAttribute(delegationtype.AttributeKeyAmount, r.Amount.String()),
+			sdk.NewAttribute(delegationtype.AttributeKeyCompletedEpochID, r.CompletedEpochIdentifier),
+			sdk.NewAttribute(delegationtype.AttributeKeyCompletedEpochNumber, fmt.Sprintf("%d", r.CompletedEpochNumber)),
+			sdk.NewAttribute(delegationtype.AttributeKeyUndelegationID, fmt.Sprintf("%d", r.UndelegationId)),
+			sdk.NewAttribute(delegationtype.AttributeKeyTxHash, params.TxHash.String()),
+			sdk.NewAttribute(delegationtype.AttributeKeyBlockNumber, fmt.Sprintf("%d", r.BlockNumber)),
+			sdk.NewAttribute(delegationtype.InstantUnbonding, "instant unbonding case"),
```
MaxMustermann2:
```suggestion
			sdk.NewAttribute(delegationtype.InstantUnbonding, fmt.Sprintf("%t", true)),
```

x/delegation/types/keys.go
```diff
@@ -30,6 +30,11 @@ const (
 
 	// ByteLengthForUint64 the type of chainID length is uint64, uint64 has 8 bytes.
 	ByteLengthForUint64 = 8
+
+	// For instant penalty
+	InstantUndelegationPenaltyKey = "instant_undelegation_penalty"
```
MaxMustermann2: Change to a common `Params` key.

MaxMustermann2 (changes requested):

x/delegation/keeper/delegation.go
```diff
@@ -132,6 +132,103 @@ func (k *Keeper) delegateTo(
...
+	if !found {
+		return errorsmod.Wrapf(delegationtype.ErrEpochIdentifierNotExist, "identifier:%s", completedEpochID)
+	}
+
+	// Apply instant penalty only when the completion epoch is more than one epoch away
+	if completedEpochNumber-epochInfo.CurrentEpoch > 1 {
+		// Get the instant undelegation penalty
+		penalty := k.GetInstantUndelegationPenalty(ctx)
+		penaltyAmount := params.OpAmount.Mul(sdk.NewInt(int64(penalty))).Quo(sdk.NewInt(100))
+		r.Amount = r.Amount.Sub(penaltyAmount)
```
TimmyExogenous:
```suggestion
```
Remove this line, as the penalty should only impact the final completed amount, and it would be equal to the difference between the submitted and actual amounts.

proto/imuachain/delegation/v1/genesis.proto
```diff
@@ -13,16 +14,18 @@
 // or in the event of a restart. At this point, it is only built with
 // the former in mind. There are no params in this module.
 message GenesisState {
+  // params refers to the parameters of the module.
+  Params params = 1 [(gogoproto.nullable) = false];
```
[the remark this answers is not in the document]
MaxMustermann2: Please update the comment of `GenesisState` as advised.

@TimmyExogenous do you think this change needs a database migration? If it's getting difficult maybe we should put it back at the bottom; apologies for the double round.

x/delegation/keeper/delegation.go
```diff
@@ -132,6 +132,102 @@ func (k *Keeper) delegateTo(
 	return nil
 }
 
+// InstantUndelegateFrom undelegates asset from operator with instant unbonding
```
MaxMustermann2, line 135: It seems that the contents of this function are mostly shared with `UndelegateFrom`. Can you make one common function instead?

x/delegation/keeper/msg_update_params.go
```diff
@@ -0,0 +1,31 @@
...
+		)
+	}
+
+	k.Logger(ctx).Info(
+		"UpdateParams request",
+		"authority", k.authority,
+		"params.Authority", req.Authority,
+	)
+
+	k.SetParams(ctx, req.Params)
```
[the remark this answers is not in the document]
MaxMustermann2: Please add this.

proto/imuachain/delegation/v1/genesis.proto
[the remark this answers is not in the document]
MaxMustermann2: I think that we will need a database migration regardless of the position of `Params` to initialize the value with 25%.

MaxMustermann2 (changes requested):

proto/imuachain/delegation/v1/genesis.proto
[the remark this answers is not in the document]
TimmyExogenous: > database

It looks like there will be a reset upgrade based on genesis.json, rather than an in-store upgrade for the Oracle, right? @leonz789  
If that's the case, we can handle all these changes entirely in this upgrade.  
Additionally, the new distribution module hasn't handled data migration either, so it might also require a reset upgrade.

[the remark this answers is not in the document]
leonz789: Yes, the oracle 2-phases feature requires a reset upgrade.
I think it’s better to include all these changes in the next version and skip version migrations before mainnet.

TimmyExogenous (approved):

MaxMustermann2 (approved):

PR #377, opened 2025-08-04
fix(operator): fix the issue regarding AVS voting power update

TimmyExogenous:
## Description

* Fix the issue related to AVS voting power updates. The operator's USD value could not be updated correctly in the following scenario:
  1. An AVS is registered with an empty asset list.
  2. The operator opts into the AVS.
  3. The asset list is updated after one epoch.
* Add an RPC and CLI command to query the operator's asset USD value.

## Summary by CodeRabbit

* **New Features**
  * Added a new API and CLI command to query the USD value of a specific operator asset for a given epoch.
* **Improvements**
  * The API endpoint for querying AVS USD value now requires the AVS address as a path parameter.
  * Enhanced error messages and validation for operator asset queries.
  * Improved consistency in field naming for AVS address parameters.
  * Added validation to prevent deregistration of AVS with positive voting power.
* **Bug Fixes**
  * Minor corrections in CLI command usage examples and field names.


MaxMustermann2 (approved):

x/operator/keeper/grpc_query.go
```diff
@@ -226,15 +226,37 @@ func (k *Keeper) QueryOperatorUSDValue(ctx context.Context, req *types.QueryOper
...
+		Amount: assetUSDValue,
+	}, nil
+}
+
 func (k *Keeper) QueryAVSUSDValue(ctx context.Context, req *types.QueryAVSUSDValueRequest) (*types.DecValueField, error) {
 	if req == nil {
 		return nil, status.Errorf(codes.InvalidArgument, "empty request")
 	}
 	c := sdk.UnwrapSDKContext(ctx)
-	if !common.IsHexAddress(req.AVSAddress) {
```
bwhour, line 234: why do we change this typo **AVSAddress** ？ it is recommended to use such _**AVSAddress**_

bwhour (approved):
LGTM.

x/operator/keeper/grpc_query.go
TimmyExogenous: There is an incompatibility when it is used with RPC.

PR #408, opened 2026-03-05
fix: correctly jail and/or tombstone

MaxMustermann2:
Many distinct but related issues are fixed by this PR
1. When a validator double signs (equivocates), slashing is not applied.
   That is because the validator's status is (incorrectly) reported as
   unbonded. Instead of tracking the status (between unbonding,
   unbonded, and bonded), we simply use a value of unspecified.
2. After a validator is tombstoned, it is possible for it to rejoin the
   validator set by changing the key. This is now forbidden by creating
   an operator frozen state
3. If a validator changes its consensus key before the slashing evidence
   is submitted and processed, the new consensus key can be unjailed.
   This is because x/slashing, which handles jailing, indexes the
   validator's tombstone and jailing status by consensus key. Since the
   new consensus key isn't tombstoned, it can be unjailed. The fix is to
   simply copy the signing info from the old key to the new key whenever
   it is slashed.
4. A validator can escape downtime slashing by constantly changing their
   key (how often is left as an exercise for the reader). This is
   because the signing info is saved on a per-key basis, and each new
   key would start with fresh signing info (no missed blocks). To fix
   this, the operator hooks in x/dogfood copy over the signing info.
5. The debt of a consensus key being reused after the unbonding period
   was being retained. This refers primarily to the missed blocks counter.
6. Tombstoned keys could re-enter the system.

## Summary by CodeRabbit

* **New Features**
  * Operator freeze/unfreeze: new transaction, CLI command, gRPC query and REST endpoint, plus genesis support and explorer/docs entry.

* **Bug Fixes & Refactoring**
  * Frozen operators are excluded from voting power and certain operator actions (opt-in/out, edits); frozen-state checks centralized across flows.

* **Chores**
  * Removed legacy slash precompile and its CLI/proto/swagger endpoints and tests; updated API surface accordingly.


x/dogfood/keeper/impl_sdk.go
```diff
@@ -104,60 +124,175 @@ func (k Keeper) SlashWithInfractionReason(
...
+		if err := k.operatorKeeper.FreezeOperator(
+			ctx, accAddress,
+		); err != nil {
+			panic(
+				fmt.Sprintf(
+					"Logic error: failed to freeze operator %s: %v",
+					accAddress.String(), err,
+				),
+			)
+		}
```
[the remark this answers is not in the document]
MaxMustermann2: No, any operator who is frozen cannot be frozen again. There are enough guards. If this happens, it would mean that it's a mistake in the code.

x/dogfood/types/expected_keepers.go
```diff
@@ -93,3 +96,12 @@ type AVSKeeper interface {
...
 }
+
+type SlashingKeeper interface {
+	GetValidatorSigningInfo(
+		sdk.Context, sdk.ConsAddress,
+	) (slashingtypes.ValidatorSigningInfo, bool)
+	SetValidatorSigningInfo(
+		sdk.Context, sdk.ConsAddress, slashingtypes.ValidatorSigningInfo,
+	)
+}
```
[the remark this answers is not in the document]
MaxMustermann2: It builds fine, you are wrong.

x/operator/keeper/operator.go
```diff
@@ -757,3 +758,9 @@ func (k *Keeper) IsUnbondingRelatedAVS(ctx sdk.Context, avsAddr string) bool {
 	}
 	return isUnbondingRelatedAVS
 }
+
+// OperatorAssetSlashedProportion returns the slashed proportion of an operator's asset.
+func (k *Keeper) OperatorAssetSlashedProportion(ctx sdk.Context, opAddr sdk.AccAddress, assetID string, startHeight, endHeight uint64) sdkmath.LegacyDec {
+	// TODO: implement this
+	return sdkmath.LegacyZeroDec()
+}
```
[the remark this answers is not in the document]
MaxMustermann2: Removed.

x/operator/keeper/usd_value.go
```diff
@@ -168,6 +168,7 @@ func (k *Keeper) DeleteAllOperatorsUSDValueForAVS(ctx sdk.Context, avsAddr strin
 func (k *Keeper) GetOperatorOptedUSDValue(ctx sdk.Context, avsAddr, operatorAddr string) (operatortypes.OperatorOptedUSDValue, error) {
 	// return zero if the operator has opted-out of the AVS effectively
 	if k.IsOptedOutAndEffective(ctx, operatorAddr, avsAddr) {
+		fmt.Println("opted out and effective", operatorAddr)
 		return operatortypes.OperatorOptedUSDValue{
```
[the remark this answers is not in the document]
MaxMustermann2: Removed.

x/dogfood/keeper/genesis.go
```diff
@@ -64,14 +64,17 @@ func (k Keeper) InitGenesis(
...
 		// we have already checked in gs.Validate() that wrappedKey is not nil
 		wrappedKey := keytypes.NewWrappedConsKeyFromHex(val.PublicKey)
 		// check that an operator exists
-		if found, _ := k.operatorKeeper.GetOperatorAddressForChainIDAndConsAddr(
+		if found, valAddr := k.operatorKeeper.GetOperatorAddressForChainIDAndConsAddr(
 			ctx, chainIDWithoutRevision, wrappedKey.ToConsAddr(),
 		); !found {
 			panic(fmt.Sprintf("operator not found for key %s", val.PublicKey))
+		} else {
+			valAddrs = append(valAddrs, sdk.ValAddress(valAddr.Bytes()))
```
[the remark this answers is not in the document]
MaxMustermann2: Removed.

x/dogfood/keeper/impl_sdk.go
```diff
@@ -104,60 +124,175 @@
...
+		// Permanently freeze the operator globally
+		if err := k.operatorKeeper.FreezeOperator(
+			ctx, accAddress,
+		); err != nil {
+			panic(
+				fmt.Sprintf(
+					"Logic error: failed to freeze operator %s: %v",
+					accAddress.String(), err,
+				),
+			)
```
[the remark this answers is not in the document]
MaxMustermann2: Acceptable. If this panic happens, it indicates a logic error.

x/dogfood/keeper/impl_operator_hooks.go
```diff
@@ -20,14 +20,32 @@ func (k *Keeper) OperatorHooks() OperatorHooksWrapper {
...
+	ctx sdk.Context, accAddress sdk.AccAddress,
+) {
+	cc, writeFunc := ctx.CacheContext()
+	if err := h.keeper.Hooks().AfterValidatorCreated(
+		cc, sdk.ValAddress(accAddress),
+	); err != nil {
+		h.keeper.Logger(ctx).Error("error in AfterValidatorCreated", "error", err)
+	} else {
+		writeFunc()
+	}
```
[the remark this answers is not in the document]
MaxMustermann2: No, the only place where the hook is called is in x/slashing and it won't be a problem.

```go
// AfterValidatorCreated adds the address-pubkey relation when a validator is created.
func (h Hooks) AfterValidatorCreated(ctx sdk.Context, valAddr sdk.ValAddress) error {
	validator := h.k.sk.Validator(ctx, valAddr)
	consPk, err := validator.ConsPubKey()
	if err != nil {
		return err
	}

	return h.k.AddPubkey(ctx, consPk)
}
```

x/dogfood/keeper/impl_operator_hooks.go
```diff
@@ -58,6 +76,14 @@ func (h OperatorHooksWrapper) AfterOperatorKeyReplaced(
...
 		unbondingEpoch := h.keeper.GetUnbondingCompletionEpoch(ctx)
 		h.keeper.AppendConsensusAddrToPrune(ctx, unbondingEpoch, consAddr)
+		oldConsAddr := oldKey.ToConsAddr()
+		newConsAddr := newKey.ToConsAddr()
+
+		// Map standard cosmos hooks (call AfterValidatorCreated to register the new key).
+		// This creates a blank ValidatorSigningInfo internally.
+		h.afterValidatorCreated(ctx, accAddress)
+		// copy from old cons addr to new cons addr
+		h.keeper.CopyValidatorSigningInfo(ctx, oldConsAddr, newConsAddr)
```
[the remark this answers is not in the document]
MaxMustermann2: I don't understand.

x/operator/keeper/consensus_keys.go
```diff
@@ -463,6 +463,9 @@ func (k Keeper) ValidatorByConsAddrForChainID(
 		ctx.Logger().Error("ValidatorByConsAddrForChainID new validator error", "err", err)
 		return stakingtypes.Validator{}, false
 	}
+	// set this because the default is Unbonded and this module does not know the true status,
+	// which is instead stored in downstream modules.
+	val.Status = stakingtypes.Unspecified
```
TimmyExogenous, line 485: nice catch

x/dogfood/keeper/validators.go
```diff
@@ -46,62 +46,41 @@ func (k Keeper) UnbondingTime(ctx sdk.Context) time.Duration {
...
 	for _, change := range changes {
 		addr := change.Key.ToConsAddr()
 		val, found := k.GetImuachainValidator(ctx, addr)
 		switch found {
+		// same consensus key exists in our store
 		case true:
 			// update or delete an existing validator.
 			// assumption: power can not be negative.
 			if change.Power < 1 {
-				// guard for errors within the hooks.
```
TimmyExogenous, line 59: nice refinement.

x/operator/keeper/freeze.go
```diff
@@ -0,0 +1,44 @@
+package keeper
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+
+	operatortypes "github.com/imua-xyz/imuachain/x/operator/types"
+)
+
+// FreezeOperator permanently freezes an operator.
+func (k *Keeper) FreezeOperator(ctx sdk.Context, addr sdk.AccAddress) error {
```
TimmyExogenous, line 12: Should we consider a function to unfreeze the operator in the future? This function should be approved by a multisig address, just like a parameter update.

TimmyExogenous (approved):
Great work

TimmyExogenous:
https://github.com/imua-xyz/imuachain/blob/22120fa6a8d439669f6173917c4f5acbe639ee4a/x/operator/keeper/abci.go#L66
We should add a frozen check here to apply the freeze to all opted-in AVSs, right?

x/dogfood/keeper/impl_sdk.go
```diff
@@ -104,60 +124,175 @@ func (k Keeper) SlashWithInfractionReason(
...
+		} else {
+			k.CopyValidatorSigningInfo(ctx, addr, currentConsAddr)
+		}
+	}
+	if infraction == stakingtypes.Infraction_INFRACTION_DOUBLE_SIGN {
+		// Permanently freeze the operator globally
+		if err := k.operatorKeeper.FreezeOperator(
+			ctx, accAddress,
+		); err != nil {
+			panic(
```
bwhour, line 208: If the same operator double-signs at different block heights, CometBFT will submit two separate pieces of evidence. The first freezing is successful, and the second one triggers ErrOperatorAlreadyFrozen → panic → chain halt.  Only "unexpected" errors should panic, while "expected repeated freezes" can be safely ignored. otherwise the externally triggerable panics are a **DoS** vector in the chain.

x/dogfood/keeper/impl_sdk.go
```diff
@@ -104,60 +124,175 @@ func (k Keeper) SlashWithInfractionReason(
...
+		ctx, oldConsAddr,
+	)
+	if found {
+		info = slashingtypes.NewValidatorSigningInfo(
+			newConsAddr,
+			info.StartHeight,
+			info.IndexOffset,
+			info.JailedUntil,
+			info.Tombstoned || markTombstone,
+			info.MissedBlocksCounter,
```
bwhour, line 238: Downtime tracking mechanism of Cosmos SDK x/slashing: 
**HandleValidatorSignature** is called every block and maintains a sliding window of fixed size. **MissedBlockBitArray[index] r**ecords whether a block was missed at each window position, and **MissedBlocksCounter** is the sum of true values in the bit array. When the window slides, the flipping of old positions will update the counter synchronously. The counter and the bit array must be consistent; otherwise, the **counter** will drift.

copyValidatorSigningInfo only copies ValidatorSigningInfo (including MissedBlocksCounter) but does not copy MissedBlockBitArray

x/operator/keeper/consensus_keys.go
bwhour, line 425: The comment (line 406) claims that "These operators are neither jailed, nor frozen", but there is no check for "frozen" in the implementation. Add the IsOperatorFrozen check in the filtering to make the implementation consistent with the comment.

x/dogfood/keeper/impl_operator_hooks.go
```diff
@@ -58,6 +76,14 @@ func (h OperatorHooksWrapper) AfterOperatorKeyReplaced(
 		// pruning for the consensus address at the end of the unbonding completion epoch.
 		unbondingEpoch := h.keeper.GetUnbondingCompletionEpoch(ctx)
 		h.keeper.AppendConsensusAddrToPrune(ctx, unbondingEpoch, consAddr)
+		oldConsAddr := oldKey.ToConsAddr()
```
bwhour, line 112: the line 79 and line 68 are the same which calculates the same value

proto/imuachain/operator/v1/genesis.proto
```diff
@@ -44,6 +44,8 @@ message GenesisState {
   repeated OperatorAssetUSDValue operator_asset_usd_values = 9 [(gogoproto.nullable) = false];
   // params is the params of the operator module.
   Params params = 10 [(gogoproto.nullable) = false];
+  // frozen_operators is a list of frozen operators
```
bwhour, line 47: _**proto/imuachain/imslash/v1/*.proto**_ 
The module has been deleted, but the proto definition files have not been cleaned up.

x/operator/keeper/freeze.go
MaxMustermann2: Yeah that makes sense.

x/dogfood/keeper/impl_sdk.go
MaxMustermann2: Repeated evidence will be ignored by x/evidence.

```go
        // ignore if the validator is already tombstoned
	if k.slashingKeeper.IsTombstoned(ctx, consAddr) {
		logger.Info(
			"ignored equivocation; validator already tombstoned",
			"validator", consAddr,
			"infraction_height", infractionHeight,
			"infraction_time", infractionTime,
		)
		return
	}
```

x/dogfood/keeper/impl_sdk.go
```diff
@@ -104,60 +124,182 @@ func (k Keeper) SlashWithInfractionReason(
...
-	// TODO(mm): add list of assets to be slashed (and not just all of them).
-	// based on yet to be finalized slashing design.
-	return k.operatorKeeper.SlashWithInfractionReason(
+
+	slashingOldKey := false
+	var currentConsAddr sdk.ConsAddress
+	found, currentKey, err := k.operatorKeeper.GetOperatorConsKeyForChainID(
+		ctx, accAddress, chainIDWithoutRevision,
+	)
+	if err == nil {
```
bwhour, line 145: is it better to log errors and explain why the safety guarantees cannot be met, rather than silently skipping here ?

x/dogfood/keeper/impl_operator_hooks.go
```diff
@@ -56,14 +73,23 @@ func (h OperatorHooksWrapper) AfterOperatorKeyReplaced(
...
 		unbondingEpoch := h.keeper.GetUnbondingCompletionEpoch(ctx)
-		h.keeper.AppendConsensusAddrToPrune(ctx, unbondingEpoch, consAddr)
+		// nb. whenever such a pruning is complete, we can also tell x/slashing to delete
+		// the mapping from consensus address to consensus public key. this is done via the
+		// AfterValidatorRemoved hook which is called when the pruning is performed.
+		h.keeper.AppendConsensusAddrToPrune(ctx, unbondingEpoch, oldConsAddr)
+		// Map standard cosmos hooks (call AfterValidatorCreated to register the new key).
+		// This creates a blank ValidatorSigningInfo internally.
+		h.afterValidatorCreated(ctx, accAddress)
```
bwhour: Please correct me under this scenario:

> afterValidatorCreated failed (the error is logged but writeFunc is not called), the new key’s ValidatorSigningInfo was not created. Then CopyValidatorSigningInfo executes unconditionally, enters the found=false branch, initializes blank default values, loses the old key’s missed-block history, and StartHeight is reset to the current height.

MaxMustermann2: Resolved.

x/dogfood/keeper/impl_sdk.go
MaxMustermann2: Added a panic, this is a gross violation.

x/operator/client/cli/tx.go
```diff
@@ -400,3 +401,29 @@ func CmdUpdateParams() *cobra.Command {
...
+				return err
+			}
+			addr := strings.ToLower(args[0])
+			if _, err := sdk.AccAddressFromBech32(addr); err != nil {
+				return err
+			}
+			msg := &types.UnfreezeOperatorReq{
+				Authority:       clientCtx.GetFromAddress().String(),
+				OperatorAddress: addr,
+			}
```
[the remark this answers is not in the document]
MaxMustermann2: If the authority is set to an EOA, this will proceed. Do not assume that the authority will always be x/governance or an x/group multi-sig.

TimmyExogenous (approved):

MaxMustermann2:
@coderabbitai review

x/dogfood/keeper/key_change_escape_test.go
```diff
@@ -0,0 +1,993 @@
...
+		&operatortypes.SetConsKeyReq{
+			Address:       operatorAddr2.String(),
+			AvsAddress:    suite.AvsAddress,
+			PublicKeyJSON: consKeyA.ToJSON(),
+		},
+	)
+	// it should be fully rejected to take over a jailed key
+	suite.Require().Error(err)
+	suite.Require().Nil(response)
+}
```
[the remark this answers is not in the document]
MaxMustermann2: Added

x/dogfood/keeper/key_change_escape_test.go
```diff
@@ -0,0 +1,993 @@
...
+	currentlyMissed := int64(0)
+	blocksToMiss := maxMissed - currentlyMissed + 1
+	blocksToReachMin := minHeight
+	blocksToCommit := blocksToMiss
+	if blocksToCommit < blocksToReachMin {
+		blocksToCommit = blocksToReachMin
+	}
+	for i := 0; i < int(blocksToCommit); i++ {
+		suite.CommitWithInfo(validators, []int{2}, time.Nanosecond)
+	}
```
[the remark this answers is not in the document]
MaxMustermann2: Updated

x/dogfood/keeper/impl_sdk.go
```diff
@@ -104,60 +124,208 @@ func (k Keeper) SlashWithInfractionReason(
...
+		info.JailedUntil = evidencetypes.DoubleSignJailEndTime
+	case stakingtypes.Infraction_INFRACTION_DOWNTIME:
+		// reset so the operator doesn't get slashed immediately upon unjail.
+		// x/slashing does this for the old key a few lines after calling this
+		// function.
+		info.MissedBlocksCounter = 0
+		info.IndexOffset = 0
+		info.JailedUntil = ctx.BlockTime().Add(
+			k.slashingKeeper.DowntimeJailDuration(ctx),
+		)
```
bwhour, line 272: **Scenario:** If a key replacement (UNSPECIFIED copy) and an old-key downtime slash (DOWNTIME copy) happen in the same epoch, the DOWNTIME copy resets 
  MissedBlocksCounter to zero but leaves the bit array entries written by the prior UNSPECIFIED copy intact, causing the counter to drift negative and 
  raising the effective downtime slash threshold by up to maxMissed blocks. We need add the method the SDK has exposed after **Line-239**, 
> `k.slashingKeeper.ClearValidatorMissedBlockBitArray(ctx, newConsAddr)`

MaxMustermann2:
@coderabbitai review

bwhour (approved):
LGTM.

MaxMustermann2:
Important files to review in this PR

x/dogfood/keeper/abci.go -> validator removed hook is now here
x/dogfood/keeper/impl_operator_hooks.go
x/dogfood/keeper/impl_sdk.go
x/dogfood/keeper/validators.go -> validator removed hook deleted

x/operator/keeper/consensus_keys.go -> setting the validator status as unspecified instead of the default value of bonded
x/operator/keeper/freeze.go
x/operator/keeper/msg_server.go -> ability to unfreeze operator by governance

x/operator/types/genesis.go
```diff
@@ -544,6 +546,34 @@ func (gs GenesisState) ValidateOperatorAssetUSDValues(operators map[string]struc
...
+		if _, ok := operators[operator]; !ok {
+			return ErrInvalidGenesisData.Wrapf(
+				"ValidateFrozenOperators: unknown operator for the frozen operator, %s",
+				operator,
+			)
+		}
+		return nil
+	}
+	_, err := utils.CommonValidation(gs.FrozenOperators, seenValueFunc, validationFunc)
+	if err != nil {
```
leonz789, line 571: we can just return err

x/operator/keeper/consensus_keys.go
```diff
@@ -103,12 +104,21 @@ func (k *Keeper) setOperatorConsKeyForChainID(
...
 		} else {
-			k.Hooks().AfterOperatorKeySet(ctx, opAccAddr, chainID, wrappedKey)
+			if err := k.Hooks().AfterOperatorKeySet(
+				ctx, opAccAddr, chainID, wrappedKey,
+			); err != nil {
+				return err
+			}
 		}
 	}
+	writeFunc()
```
adu-web3, line 128: why do we need dedicated extra atomicity for `setOperatorConsKeyForChainID`?  wouldn't the tx be reverted on any error returned by it?

leonz789 (approved):

adu-web3:
I have worked with the agent to cover most of the important changes brought by this PR, and two possible findings are found and fixed, together with some newly added invariant tests, and I have committed them. But they could be false positive, so you might need to review the commits and decide if the finding is true, as well as if the fix is proper. 

```
We found one concrete bug pattern: during slash handling, the code
   could still copy/migrate validator signing state to the new
  consensus key even when the underlying operator slash operation
  failed (or did not really execute). In plain terms, one subsystem
  (slashing state) could say “punishment/migration happened,” while
  another subsystem (operator slash state) says “slash did not
  commit.” That mismatch is dangerous because later logic reads both
   stores and assumes they represent the same event history. Root
  cause: this PR area sits across a multi-step lifecycle (old key ->
   new key -> unbonding window -> prune) and multiple modules
  (dogfood, operator, slashing, evidence/oracle), so correctness
  depends on strict ordering and success-coupling, not just
  individual function behavior. We fixed this by only copying
  signing info after a confirmed successful slash path, keeping
  double-sign freeze as a safety action, and making freeze
  idempotent so repeated double-sign inputs do not panic. We then
  added focused invariants/tests to lock the behavior at key
  boundaries (same-epoch rotations, stale-key slash, prune/no-op
  slash, frozen unjail, and randomized lifecycle transitions).
  ```

File: x/dogfood/keeper/impl_sdk.go [hunk 24]
```diff
@@ -104,60 +128,252 @@ func (k Keeper) SlashWithInfractionReason(
...
+			"operator AVS slash failed (dogfood StakingKeeper)",
+			"error", slashErr,
+			"operator", accAddress.String(),
+			"cons_addr", addr.String(),
+			"infraction", infraction.String(),
+		)
+	}
+	// Only migrate x/slashing state when the economic slash committed; otherwise Copy could write
+	// inconsistent signing info relative to operator state (e.g. veto / SlashAssets failure).
+	if slashingOldKey && attempted && slashErr == nil {
```
MaxMustermann2: The presence of `attempted` here and the check `slashFactor.IsZero()` in the called function is problematic. If the slash factor is set to 0 for downtime (which in our current economic model is), a validator could escape downtime jailing through this route.

MaxMustermann2: 2. The condition `slashErr == nil` is also not necessary here. Regardless of whether a validator is economically slashed, they should be jailed or tombstoned as applicable. This is performed by the SDK on the old key, and we apply it here on the new key.

MaxMustermann2: 3. The `attempted` variable can be `false` in two more cases: (a) a `nil` or `negative` slash factor, which are guarded against in `x/slashing` parameter vallidation functions, and (b) a missing AVS-by-chain-id, which is guarded in the call to `GetOperatorAddressForChainIDAndConsAddr` above. Thus, it must be `true` and the condition is unwarranted.

x/dogfood/keeper/impl_sdk.go
```diff
@@ -104,60 +128,252 @@ func (k Keeper) SlashWithInfractionReason(
...
+		// copy from old key to new key
+		// the slashing keeper tombstones after calling this function. hence,
+		// copy directly will not succeed. we must do it ourselves.
+		k.CopyValidatorSigningInfo(
+			ctx, addr, currentConsAddr, infraction,
+		)
+	}
+	if infraction == stakingtypes.Infraction_INFRACTION_DOUBLE_SIGN && power > 0 &&
```
MaxMustermann2: `power > 0` is guaranteed by Tendermint and this is a Tendermint interface. So that condition is not required.

x/dogfood/keeper/impl_sdk.go
```diff
@@ -104,60 +128,252 @@ func (k Keeper) SlashWithInfractionReason(
...
+		(attempted || slashErr != nil) {
```
MaxMustermann2: 1. `slashErr` can only be non-nil if `attempted` is true. So we can drop `attempted` from this condition.
2. The freeze should be applied if `slashErr` is true or false. This is written down immediately below this line as a comment: `freeze even if the economic slash path failed`.

x/dogfood/keeper/impl_sdk.go
```diff
@@ -104,60 +128,252 @@ func (k Keeper) SlashWithInfractionReason(
...
+		// Double-sign is a consensus safety fault: freeze even if the economic slash path failed
+		// (so veto or asset errors cannot leave the operator unfrozen).
+		if k.operatorKeeper.IsOperatorFrozen(ctx, accAddress) {
```
MaxMustermann2: This change I am more willing to keep. While it is known that the in-built AVS's codebase (specifically `x/evidence`) refuses to tombstone the same validator twice, the other AVSs may not provide the same guarantees. So it is not necessary to panic if an operator is already frozen and a request is made to freeze it.

x/operator/keeper/consensus_keys.go
MaxMustermann2: Added because of function complexity; no particular rationale

MaxMustermann2:
@coderabbitai review

